PR #1, opened 2022-05-09
Move from exceptions to tag clearing

PeterRugg:
As previously discussed at length, move towards tag clearing on non-monotonic register-register operations in RISC-V and suggest it more strongly in general.

Currently this is just an initial cut at some text changes, which should likely be merged alongside the sail changes (https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/56) once validated in hardware.

### DIFF
--- a/chap-architecture.tex
+++ b/chap-architecture.tex
@@ -2056,7 +2056,7 @@ CHERI introduces several new potential failure modes:
   favored throwing an exception at the earliest possible point: the
   instruction attempting to violate guarded manipulation.
 
-  Another potential design choice is to instead strip the tag from the value
+  However, in all current architectures, we instead strip the tag from the value
   being written back to a target capability register, which maintains our
   security safety properties, but defers exception delivery until an attempted
   dereference -- e.g., an instruction load via the resulting invalid
@@ -2067,6 +2067,8 @@ CHERI introduces several new potential failure modes:
   exception delivery means that failures that could otherwise be easily
   detected and handled by a compiler or language run time via an explicit tag
   check are now complex to handle.
+  In addition, stripping the tag avoids encouraging implementations that
+  are vulnerable to speculative side channel attacks.
 %%%% STRANGE USAGE.  more difficult to handle?  more complex?
 
   When using tag stripping in ISAs with status registers (e.g., ARMv8-A), the
@@ -2075,11 +2077,6 @@ CHERI introduces several new potential failure modes:
   For ISAs without status registers, checking results can come at a
   significant cost, and a deferred exception delivery at time of dereference
   will be the best choice for performance-critical code.
-
-  We therefore make design choices about exception delivery for violations of
-  guarded manipulation in a case-by-case basis, taking into account more
-  general architectural design philosophies, and also specific use cases where
-  software may benefit from tag clearing rather than exception delivery.
 \end{description}
 
 %\subsection{Object-Capability Invocation}
@@ -2349,10 +2346,6 @@ For example:
 \item New capability-relative load and store instructions will trap if they
   attempt to access memory locations in a manner not authorized by the
   explicitly presented capability.
-
-\item New capability-manipulation instructions may trap if they violate
-  guarded-manipulation rules, such as by attempting to increase the bounds on
-  a capability.
 \end{itemize}
 
 \noindent
--- a/chap-architecture.tex
+++ b/chap-architecture.tex
@@ -2224,7 +2224,7 @@ In operating-system design, {\em privileges} are a special set of rights
 exempting a component from the normal protection and access-control models --
 perhaps for the purposes of system bootstrapping, system management, or
 low-level functionality such as direct hardware access.
-In CHERI, three notions of privilege are defined, complementing current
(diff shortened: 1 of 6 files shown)


chap-model.tex
```diff
@@ -1087,17 +1086,46 @@ \subsection{Failure Modes and Exceptions}
...
+when the operation would be illegal, or risk being vulnerable to
+denial-of-service attacks.
+This may require it to recreate the hardware-performed checks in software.
+
+With a deferred-exception approach, as well as avoiding these issues,
+microarchitecture is simplified by reducing the set of instructions that can
+throw exceptions.
+While it is initially tempting to delay performing the required checks,
+forwarding the common-case value and later flushing the pipeline if a check
+fails, this leads to exploitable speculative speculative side channel
```
francislaus, line 1119: duplicate word "speculative"

chap-model.tex
```diff
@@ -1087,17 +1086,45 @@ \subsection{Failure Modes and Exceptions}
...
+of a capability's authority, instead clearing the tag on operations that
+would otherwise violate monotonicity.
+
+\pdrnote{The following two paragraphs perhaps belong in rationale?}
+The early exception approach offers slightly improved debuggability
+(by exposing the error earlier).
+However, early exceptions limit compiler optimization as instructions that may
+throw exceptions are restricted in how they can safely be reordered.
+For example, a bounds restriction performed within a loop could be hoisted to
+before the loop, except for the possibility that the bounds are being widened.
```
arichardson, line 1108: Maybe something more detailed like this:

```suggestion
For example, this prevents a bounds restrictions performed within a loop from
being hoisted outside the loop, unless that instructions is always executed.
If the loop is not always entered, this could turn a conditional execution of a  trapping instruction into an unconditional one.\footnote{%
This is not just a theoretical possibility -- we observed this happening in the FreeBSD kernel and had to modify the compiler to avoid hoisting any
potentially-trapping CHERI instructions.
}
```

arichardson (approved):
LGTM once the sail changes have been merged.

rwatson:
Per comments on https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/56, you can likely GC some exception cause definitions.

tariqkurd-repo:
This PR has been open since May, when do you expect it to be merged? I believe that you're waiting for the SAIL model to be updated to the new spec.

PeterRugg:
@tariqkurd-repo Soon! The sail PR is ready to go (https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/56). I've been doing all the implementations (Flute is done, just finishing Toooba). Once I confirm that modified Toooba boots CheriBSD, I think we can do the switch. Hopefully this week.

tariqkurd-repo:
That's great, thanks @PeterRugg . I look forwarding to seeing the updated spec.

PeterRugg:
> I do think that the approach we should take is to only document tag clearing in the model chapter and instead add a separate section over in the rationale chapter that discusses the other approach and the reasons we moved away from it. In particular, I think we want the initial architecture-neutral chapters to be as lean as possible so that someone new to CHERI is just reading what the current spec is. If there are A vs B questions that are still open we should keep them in the initial chapters, but when we've made a change going forward (such as split vs merged register files) we should just document what we now do and describe the other approach in the rationale chapter.

Yep, I'm doing a pass now finding all references to it. I'll perhaps be a bit more aggressive in pruning/moving trade-off text.

arichardson (approved):

PeterRugg:
Okay, I finally finished a pass trying to find all the text that mentioned exceptions on guarded manipulation failures. I could well have missed some though.

chap-architecture.tex
```diff
@@ -3192,40 +3192,6 @@ \section{Potential Future Changes to the CHERI Architecture}
...
-  infinity value or an exception being thrown.
-
-  A cheap tag assertion instruction that can trigger a trap when a tag is lost
-  would allow special compilation modes to improve debuggability by detecting
-  unexpected tag loss sooner.
-
-  If MIPS had a user status register, a tag-loss bit could be set implicitly
-  on tag clear, allowing intermittent conditional-branch instructions to detect
-  and handle loss.
- \jhbnote{This all seems OBE by moving to tag-clearing semantics}
```
bsdjhb, line 3230: Thanks :)

bsdjhb (approved):
This generally looks good to me.

bsdjhb:
BTW, as a followup change after this is merged, I would propose that we make the CHERI exception code tables arch-specific (Morello already follows its own thing with separate exceptions rather than an exception code).  For x86 I will likely renumber the values into a single dense encoding omitting the RISC-V specific unaligned base exception.  I suspect for RISC-V we might also want to consider renumbering, or at the least that any variant adopted upstream by RISC-V probably should have renumbered values?

PeterRugg:
Yes, it did strike me as odd that these were architecture-neutral.

bsdjhb:
Still looks good to me.  The x86 chapter has a reference about how it tag clears unlike RISC-V that you might want to update as well (or I can take care of that after this lands if you prefer)

arichardson (approved):

PR #4, opened 2022-07-27
Permit building both normal and tech-report formats.

bsdjhb:
Add a cheri-architecture-tr.tex that defines an empty 'trformat'
command and then includes cheri-architecture.tex.  Modify
cheri-architecture.tex to use \ifdefined\trformat around blocks
specific to tech-report formatting.

With these changes, a tech-report formatted can always be built as
cheri-architecture-tr.pdf.  The default build only builds the normal
format.

### DIFF
--- a/cheri-architecture-tr.tex
+++ b/cheri-architecture-tr.tex
@@ -1,0 +1,5 @@
+%%%% Enable University of Cambridge tech-report formatting
+%%%% Use \ifdefined\trformat in other files to check the mode
+\def\trformat{}
+
+\input{cheri-architecture}
--- a/cheri-architecture.tex
+++ b/cheri-architecture.tex
@@ -1,19 +1,17 @@
-%%%% University of Cambridge tech-report formatting; enable when producing
-%%%% tech-report versions of these documents; otherwise, disable.
-%\documentclass[12pt,twoside,openright,a4paper]{report}
-%\setlength{\oddsidemargin}{-0.4mm} % 25 mm left margin
-%\setlength{\evensidemargin}{\oddsidemargin}
-%\setlength{\textwidth}{160mm}      % 25 mm right margin
-%\setlength{\topmargin}{-5.4mm}     % 20 mm top margin
-%\setlength{\headheight}{5mm}
-%\setlength{\headsep}{5mm}
-%\setlength{\footskip}{10mm}
-%\setlength{\textheight}{237mm}     % 20 mm bottom margin
-%%%% .. or regular document
+\ifdefined\trformat
+\documentclass[12pt,twoside,openright,a4paper]{report}
+\setlength{\oddsidemargin}{-0.4mm} % 25 mm left margin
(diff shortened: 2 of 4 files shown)


cheri-architecture.tex
```diff
@@ -56,6 +54,8 @@
 }
 
 %% CL tech-report format provides its own cover page
+\ifdefined\trformat
+\else
```
bsdjhb, line 58: There isn't a \ifundefined in TeX.  There is a \@ifundefined in LaTeX but the first time you use it defines its argument, hence using this construct with an empty \if body.

cheri-architecture-tr.tex
```diff
@@ -0,0 +1,5 @@
+%%%% Enable University of Cambridge tech-report formatting
+%%%% Use \ifdefined\trformat in other files to check the mode
+\def\trformat{}
+
+\input{cheri-architecture}
```
arichardson, line 5: For my thesis I used latexmk's pretex flag: `-pretex='\\newcommand*{\\buildtechreport}{yes}'`, but I guess adding a separate file might be simpler.

bsdjhb: Ah, that's one of the options I didn't find on stack exchange.  I do think the separate file makes the Makefile change simpler perhaps?  However, it is true that we could use 'make USE_TR=yes' or some such and add this to the flags passed to latexmk if that is defined?  (The other approaches were not quite as straightforward to use with a make variable like USE_TR which is why I ended up going with a separate file.)  I don't have a strong preference either way.

cheri-architecture.tex
```diff
@@ -1,19 +1,17 @@
...
+\setlength{\evensidemargin}{\oddsidemargin}
+\setlength{\textwidth}{160mm}      % 25 mm right margin
+\setlength{\topmargin}{-5.4mm}     % 20 mm top margin
+\setlength{\headheight}{5mm}
+\setlength{\headsep}{5mm}
+\setlength{\footskip}{10mm}
+\setlength{\textheight}{237mm}     % 20 mm bottom margin
+\else
 \documentclass[12pt,letterpaper,twoside,openright,fleqn]{report}
 \usepackage{etex}
```
jrtc27, line 13: I'm surprised this is a TR-vs-regular thing. cheri-architecture 97eaba24f17f5ebff1ea0c9da40b383e69c5aedd added it here (well, back then this block was in preamble.tex) as needed with tikz for some reason, but for the ISAv8 TR this was indeed commented out. I suspect this is because since ~2015 LaTeX has enabled the eTeX functionality itself so it only matters if building on an old version (perhaps Ubuntu 16.04 LTS (Xenial Xerus) didn't quite have the feature in its TeX Live distribution or Wes was using Debian 8 "Jessie" at the time).

bsdjhb: Do you mean that we should always be using the etex package?  I think if we want to move this, we should probably do it as a separate followup commit from this change.

PR #7, opened 2022-08-20
Reserve RISC-V C{Get,Set}High instructions

nwf:
See https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/55 and https://github.com/CTSRD-CHERI/cheri-architecture/pull/78#discussion_r719528101

### DIFF
--- a/def-riscv-insns.tex
+++ b/def-riscv-insns.tex
@@ -26,6 +26,11 @@
 \rvcherisrcdest[name=CGetFlags]{7}{rd}{cs1}
 \rvcherisrcdest[name=CGetAddr]{F}{rd}{cs1}
 
+% Mark as noref and reserved until landing of
+% https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/55
+\rvcherisrcdest[name=CGetHigh,noref,tablesuffix=\rvcherireservedfootnotemark]{17}{cd}{cs1}
+\rvcherisrcsrcdest[name=CSetHigh,noref,tablesuffix=\rvcherireservedfootnotemark]{16}{cd}{cs1}{rs1}
+
 \rvcherisrcsrcdest[name=CSeal]{B}{cd}{cs1}{cs2}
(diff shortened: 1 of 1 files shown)


jrtc27:
I still don't think the low ones have value, those are really just cget/setaddr even in an MTE world.

jrtc27:
(i.e., happy to reserve space for the high bits, but not for the low bits)

nwf:
I'm still not convinced about `CGetLow` == `CGetAddr`, but anyway, yes, we can separate that discussion from `C{Get,Set}High`, so this PR is now just for the latter two opcodes.

nwf:
For posterity, https://github.com/CTSRD-CHERI/cheri-specification/wiki/Tracking-discussion-for-CGetAddr---CToPtr---as-integer-alias---CGetLow

arichardson (approved):
Thanks! I'm not sure if we should stick to using the previously chosen opcodes or just allocate the next free instruction, but I'm happy with this either way.

def-riscv-insns.tex
```diff
@@ -26,6 +26,11 @@
 \rvcherisrcdest[name=CGetFlags]{7}{rd}{cs1}
 \rvcherisrcdest[name=CGetAddr]{F}{rd}{cs1}
 
+% Mark as noref and reserved until landing of
+% https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/55
+\rvcherisrcdest[name=CGetHigh,noref,tablesuffix=\rvcherireservedfootnotemark]{1C}{cd}{cs1}
```
jrtc27, line 31: I don't understand this encoding, it seems to be away from everything in the middle of nowhere (whilst CSetHigh makes sense as the encoding after CRelocate)

nwf: That's fair; indeed 1C and 1D were going to be their own new island for these accessors.  How about 17 instead?

arichardson (approved):

PR #8, opened 2022-08-24
Reserve an opcode for CGetTop

arichardson:
See https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/59

### DIFF
--- a/def-riscv-insns.tex
+++ b/def-riscv-insns.tex
@@ -27,6 +27,10 @@
 \rvcherisrcdest[name=CGetAddr]{F}{rd}{cs1}
 
 % Mark as noref and reserved until landing of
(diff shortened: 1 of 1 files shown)


nwf (approved):
Seems fine by me; kind of surprised we didn't have it already.

rmn30:
Seems fine. Do you have a specific use case in mind? Would there be any use for a `CInBounds` instruction? If you wanted to implement that in software using `CGetTop` / `CGetLen` you'd have to work around the nasty max length issue.

arichardson:
We already have a few uses of `cheri_gettop()` in cheribsd (although those don't handle the max corner case). The cheri_is_address_inbounds() check also uses this (again not handling the max case). I think adding this instruction is useful even if we add CInBounds as it is so simple and already being used (and Morello also includes it).

PR #10, opened 2022-09-07
Remove split register files from RISC-V.

bsdjhb:
These are follow-on changes to the de-MIPS PR (which mostly removed split register files from the rest of the ISA spec) to specifically remove them except as a historical note on CHERI-RISC-V.  Right now the base is the demips branch, but this will be rebased onto main once the de-MIPS changes land.


bsdjhb:
This is also perhaps the most aggressive variant of removing split register files as it removes xCCSR entirely rather than just the dirty and enable bits.

chap-cheri-riscv.tex
```diff
@@ -536,39 +474,6 @@ \subsubsection{Capability Exception Reporting}
...
-\pmnote{WIRI has been removed from the RISC-V specification.  The spec
-  now typically uses WPRI (Write-Preserve, Read-Ignore) for such
-  undefined bits (e.g. in \texttt{mstatus}).  Both WARL (Write-Any,
-  Read-Legal) and WLRL (Write-Legal, Read-Legal) seem unsuitable here.
-  The RISC-V spec unsatisfyingly does not specify what happens if
-  software doesn't preserve values for WPRI, but then this spec
-  doesn't need to either.}
-
-\begin{description}
-\item [e] The \texttt{e} ``enable'' bit \ajnote{read only? / only wriFable in u-mode? only writable with a specific permission in pcc?} tells whether capability extensions are enabled or disabled.
```
arichardson, line 566: I think keeping the enabled bit makes sense. I agree that dirty is not that useful.

sail_latex_riscv/commands.tex
```diff
@@ -7712,23 +7712,6 @@ \subsection*{Exceptions}
 
 }{\lstinputlisting[language=sail]{sail_latex_riscv/fclCMovezexecute33a689e3a631b9b905b85461d3814943.tex}}}}
 
-\newcommand{\sailRISCVfclClearexecute}{\saildoclabelled{sailRISCVfclClearzexecute}{\saildocfcl{Integer registers 8 $\times$ \emph{q} $+$ \emph{i} are each set to 0 if the \emph{i}th bit
```
arichardson, line 7715: This should probably be done by re-generating form an updated sail model - but I guess the result will be the same, so should be fine this way too.

arichardson (commented):
LGTM other than the comment on e bits.

sail_latex_riscv/commands.tex
jrtc27: Please don't touch sail_latex_riscv, just don't use the bits you don't want any more...

chap-cheri-riscv.tex
```diff
@@ -1448,3 +1343,62 @@ \subsection{\PCC{} alignment}
...
+capability register file to hold capabilities (a ``split'' register
+file).  This second approach mirrored the approach used on CHERI-MIPS.
+
+Most of the ISA is agnostic to this choice due to the principle of
+intentional-use such as the explicit use of capability-relative
+addresses versus integer pointers relative to \DDC{} or \PCC{}.
+However, the ``split'' register file approach did require a few
+additional changes:
+
+\begin{itemize}
```
jrtc27, line 1368: CSub and CGetAddr are also not needed in a merged world, though currently we provide them

bsdjhb: Hmm, I can document them here as still existing, but that is true, and I think the x86 sketch doesn't provide either of these.

bsdjhb: This does open the question of if we want to deprecate these instructions?  Clear I Think wasn't actually implemented on anything so it is simpler to just outright remove.

chap-cheri-riscv.tex
bsdjhb: Jess noted that if all we want is per-mode enable bits then the more idiomatic approach on RISC-V now would be to allocate new bits in `Xstateen` via the now-ratified `Sstateen` extension instead of using custom CSRs.  Given that we haven't yet implemented the enable bit, perhaps it would be best to add that as a replacement instead?

arichardson: That sounds good to me.

bsdjhb:
This version moves the enable bit to `[sm]stateen0`.  I added an internal note though about some of the details that aren't yet specified (and weren't before) about the semantics we expect to happen if CHERI is disabled that I'm not fully sure of.

chap-cheri-riscv.tex
nwf: `CGetAddr` goes away only if we assume that merged register files hold the address (and only the address) in their "integer subregisters".  See https://github.com/CTSRD-CHERI/cheri-specification/wiki/Tracking-discussion-for-CGetAddr-vs-CToPtr-vs-as-integer-alias-vs-CGetLow for more discussion. 

`CSub` can probably go away in the merged world regardless of layout issues.

bsdjhb: True, but we currently don't yet consider merged register files where that isn't true.  Morello still follows the TBI model.  I'm not sure if the MTE-like thing for CHERI-RISC-V also follows that model?  Regardless, if we decide to amend the spec to permit registers where CGetAddr isn't equivalent to CGetLow, that change can update this text.

rmn30: > I'm not sure if the MTE-like thing for CHERI-RISC-V also follows that model?

The version I have stole some otype bits to store a 4-bit version. I didn't want to get into the complication of using address bits for this, although it might be preferable in the long run.

bsdjhb: I meant to reply again to this yesterday.  Maybe we could soften the language around `CGetAddr` to suggest that it is not currently needed but might be needed if we stored additional data other than the address in the aliased integer register in the future?

chap-cheri-riscv.tex
bsdjhb: Hmmm, if we stored non-address bits in the lower word we would also still need `CSub`?

bsdjhb: (Assuming we don't want to follow the TBI model where code is aware of TBI and has to mask out the bits explicitly from addresses)

jrtc27: Yes. I continue to believe putting bits that aren't exposed to C as the address in the lower word is a really bad idea for all these kinds of reasons.

jrtc27: (i.e. people are free to investigate these if they want to, but we shouldn't concern ourselves with all that weirdness in the main CHERI-RISC-V spec outside of some appendix rambling to act as prior art)

bsdjhb: I've added a sentence to both `CGetAddr` and `CSub`.  I'm not sure how you wouldn't avoid the need for a `CSub` if you stored the version in the address similar to TBI but wanted to avoid explicit masking.  Were you assuming you'd do a normal `Sub` between two results of `CGetAddr`?  Presumably we wouldn't have needed `CSub` at all even in the split register case if that was reasonably efficient?

jrtc27: I guess within the same object it cancels out, and subtraction between objects is UB...

chap-cheri-riscv.tex
bsdjhb: (Replying over here from the other sub-thread): I've dropped the language update for `CSub`, but have kept the note about possibly needing `CGetAddr` in the future still.  @nwf is this updated version better?

nwf: Seems fine.  I think `CSub` needs a similar note, though, for the case that there's more than just the address in the aliased integer sub-register:
- I think we'd want it that `CSub` really does just sign-extend the subtraction of the addresses while `Sub` just projects the lower words and subtracts those, but
- I guess there's the alternative where `sub` is tag-dependent and subtracting two registers with both tags set is different than anything else

bsdjhb: I originally had a note for `CSub`, but @jrtc27 pointed out that pointer subtraction is only well-defined for subtracting between pointers to the same object.  Those pointers would have the same MTE-type color, so plain `Sub` would still work fine.  Based on that reason I removed the comment from `CSub`.  I don't think `Sub` should be tag-dependent as we've generally avoided having tag-dependent behavior like this.

nwf: OK, that's probably fair as far as the current architecture document goes.  As a completionist nit, tho', it's possible that there could be other things than MTE colors in the non-address bits in the lower word; additional permission bits, for example, and those _are_, one assumes, allowed to vary between pointers to the same object.

I'm happy for there to not be a note on `CSub`; we'll set that bridge alight when/if we come to it.

jrtc27: You can have notes about this kind of stuff if you want to but please keep it out of the main spec chapters (or tacked on at the end of them) because they're completely unexplored, likely to break various things throughout the software stack and add confusion. We really do not want people thinking it's a core part of CHERI, or seeing it and running with the idea before we've even figured out whether it makes any sense.

bsdjhb: I think what probably makes sense for the time being then perhaps is that we should have a section in the experimental chapter that talks about storing non-address bits in the low bits and what the impacts that would be in more of an architecture-neutral way.  That probably belongs as a separate thing from these changes however (and it would also be relevant for at least x86, so better to discuss it over there).

chap-cheri-riscv.tex
```diff
@@ -536,38 +476,26 @@ \subsubsection{Capability Exception Reporting}
...
-\end{description}
+\subsubsection{CHERI Extension Control}
+
+For processors supporting the \texttt{Smstateen} extension, bit 2 of
+\mstateen{} and \sstateen{} is used to enable CHERI.  When CHERI is
+disabled, attempting to execute CHERI-specific instructions will raise
+an illegal instruction fault and existing instructions will always execute in
+integer encoding mode.
+
+\jhbnote{I'm not sure what to do about \PCC and \DDC when CHERI is
```
jrtc27, line 487: IMO they continue to take full effect, and xEPCC vs xEPC should be done based on the target's state enabled bit, not the source's. That can include encoding mode too; ask for stupid things get stupid outcomes, and you can technically write code that runs in capability mode without using capability state (so long as you avoid AUIPCC), it's just a stupid thing to do. My main justification for allowing it is that it's simplifying for the architecture and microarchitecture and more consistent.

chap-cheri-riscv.tex
```diff
@@ -1448,3 +1366,71 @@ \subsection{\PCC{} alignment}
...
+  idx} fields to \xtval{}.
+
+\subsection{MMU Capability Exceptions}
+
+Initially CHERI-RISC-V reported MMU tag permission violations via the
+\riscvcheriexception{} exception code.  Unique values were assigned
+for these violations to the \texttt{cause} field of \xtval{}.
+However, storing the faulting address in \xtval{} similar to other MMU
+faults was found to be more useful.  As a result, additional top-level
+ecxeption codes (\riscvloadcappagefault{} and
```
jrtc27, line 1434:
```suggestion
exception codes (\riscvloadcappagefault{} and
```

chap-rationale.tex
```diff
@@ -260,9 +260,8 @@ \section{Capability Register File}
 might be quite different; effectively pinning the two to one another would
 reduce the efficiency of both.
 
-In CHERI-RISC-V (Chapter~\ref{chap:cheri-riscv}), we parameterize the
-instruction set to support both split and merged register files.
-Both Morello and CHERI-x86-64 use merged register files.
+CHERI-RISC-V (Chapter~\ref{chap:cheri-riscv}),
+Morello, and CHERI-x86-64 use merged register files.
```
jrtc27, line 264: Strange we don't have a ref here for x86

bsdjhb:
I've reworked this a bit for now to leave xCCSR and the enable bit in place.

jrtc27 (approved):

PR #13, opened 2022-09-15
Candidate CHERI ISA spec CC BY 4.0 license change; properly acknowledge RISC-V and CHERI-RISC-V Sail components

rwatson:


rwatson:
It looks like the Sail export to .tex for CHERI-RISC-V doesn’t include an exported license.tex that can then be included in the technical report. Probably this is a bug/gap in how the tex export happens that the copyright/license information isn’t propagated, and it should be fixed on the export side, then on the import side? (We can do something simply in the short term, but ideally the license would go with the export?).  (Tagging @rmn30 since he was pondering this also.)

jrtc27:
Where would it get the license from? It's just comments at the tops of files, and files with arbitrary names at the roots of the repositories (plural, because sail-riscv is a submodule).

rwatson:
In the worst case, perhaps some encapsulation of https://github.com/CTSRD-CHERI/sail-cheri-riscv/blob/master/LICENCE in minimal LaTeX (\verbatim?), but perhaps ideally in a more structured and/or better way.

jrtc27:
But then what about sail-riscv's LICENCE? (which has a different set of authors as it's seen more external contribution)

rwatson:
I’m presume that’s a merge concern for sail-cheri-riscv -- i.e., if it is merged forward to a more recent sail-riscv, then it should inherit updates from the sail-riscv copyright/license information, since it becomes a derived work relative to the more recent upstream?

jrtc27:
For our use cases though we don't actually need the sail-riscv license information, since we're only including sail-cheri-riscv code snippets (and the signatures for some APIs they use that happen to be in sail-riscv, I guess?), but the generated LaTeX definitely has them available.

rwatson:
And we are, in effect, distributing that content, so ideally the licenses would always have been exported with the Sail.

jrtc27:
The source is, yes, just not the PDF. The trouble is Sail has no notion of repos or projects or any meaning assigned to directories, it's just given a bunch of files to go concatenate and compile. Probably the thing that calls Sail to spit out the LaTeX should be the thing that spits out licenses, as it knows the structure of what it's building.

rmn30:
> Probably the thing that calls Sail to spit out the LaTeX should be the thing that spits out licenses, as it knows the structure of what it's building.

Indeed, but any structure that exists is lost in the translation to latex because all functions, types, etc are spat out in one directory :-(

rmn30:
Arguably the different sail code repos should be namespacing themselves with prefixes on everything...

rwatson:
Tagging @PeterSewell to catch his attention: The Sail .tex exports aren’t emitting copyright and license information for use by including documents, which we will want to fix, permitting the CHERI specification to properly attribute that work.

PeterSewell:
cc'ing Alasdair+Brian to see what they think.    Off the top of my head,
though, I'm not sure Sail-the-language or Sail-the-tool should be doing
anything for this.  As far as the language is concerned, copyright and
license info is just comments in files; it doesn't know about them -
similar to many (pretty much all?) other languages.  I can imagine various
more-or-less elaborate machinery we could invent, to make it linguistic and
automatically track it through from input files to output files (properly,
including theorem-prover and C/OCaml, not just tex), but that seems like a
lot of invention and engineering, for something that perhaps should just be
done manually.  What do RISC-V extension source files typically have?

p

nwf:
I think I agree that _Sail the language/tool_ shouldn't need to know anything about this, especially since it does not, AFAIK, have `#include`-like directives and relies on external tooling to enumerate the files to be processed.

Possibly the correct thing to do for now is to teach the `sail-cheri-riscv-latex` make target to copy `$(SAIL_CHERI_RISCV_DIR)/LICENSE` and `$(SAIL_CHERI_RISCV_DIR)/sail-riscv/LICENSE` into `$(SAIL_LATEX_RISCV_DIR)` and then just `\lstinputlisting{}` them with appropriate surrounding verbiage?

nwf:
Something like https://github.com/nwf/cheri-specification/tree/licenses-from-sail .  The generated document can be seen at https://github.com/nwf/cheri-specification/actions/runs/3077355357

jrtc27:
Sail does have `$include` FWIW

nwf:
> Sail does have `$include` FWIW

TIL!  Does it have a way to enumerate the files it included?  Does it do path search and, if so, can it enumerate the files it _would have_ included, if only they had existed?

PeterSewell:
for the tex, yes, but for the prover outputs, they might depend on multiple
files - and perhaps annoying to track which.

bacam:
Using the `$include` information probably isn't terribly helpful, because it will also include the Sail standard library files even though you're unlikely to use any snippets from there in the document.  I think the manual addition of licence files to the build system sketched above seems like a good solution.

That said, if you'd like to be able to ask Sail to generate a LaTeX command corresponding to some markdown or plain text in a comment (without tying it to a Sail definition), then it should be pretty easy to add - we already have most of the necessary pieces.

rmn30:
>  because it will also include the Sail standard library files even though you're unlikely to use any snippets from there in the document. 

There are a few val specs and doc comments, actually, because the spec includes descriptions of the standard library functions it uses.

> That said, if you'd like to be able to ask Sail to generate a LaTeX command corresponding to some markdown or plain text in a comment (without tying it to a Sail definition), then it should be pretty easy to add - we already have most of the necessary pieces.

I had a go at something like this. I added a call to a macro `\sailpath` with path of the source file at the end of each generated tex snippet.

rwatson:
I have updated the pull request to separately include licenses from both RISC-V and CHERI-RISC-V Sail models, hopefully synchronised with what we have in the spec. Would be good to get explicit head nods from involved parties that this meets their needs, in the absence of the Sail parts spitting out TeX copyright/license bits when they generate TeX versions of the specifications themselves. Should they learn to do that, we will prefer to use that output. Thanks!

rmn30:
Looks fine to me. You need to rebase on top of 2f47c16d9a0da6426fbc760ac2d4850eb0c5bbbf to fix CI failures.

nwf:
Seems fine to me; we can figure out how to not need to keep these copies up to date later.

rwatson:
> Looks fine to me. You need to rebase on top of [2f47c16](https://github.com/CTSRD-CHERI/cheri-specification/commit/2f47c16d9a0da6426fbc760ac2d4850eb0c5bbbf) to fix CI failures.

Should now be up-to-date.

PeterSewell (commented):
LGTM

PR #16, opened 2022-10-05
Qualify 32-bit vs 64-bit CHERI-RISC-V maturity.

bsdjhb:
This is in response to an open comment @arichardson had on the earlier de-MIPS PR, but this paragraph was also missed in that PR and needs updating one way or another.

### DIFF
--- a/chap-architecture.tex
+++ b/chap-architecture.tex
@@ -36,8 +36,9 @@ Our current instantiations within concrete ISAs are:
   implementations, adaptations of our CheriBSD and CheriFreeRTOS operating
   systems, Clang/LLVM/LLD toolchain, GDB debugger, and application suite.
 
-  We expect continuing disruptive modification to this ISA mapping, including
-  reencoding of many key instructions, as it transitions to a more mature
+  We aim to propose 64-bit CHERI-RISC-V as a RISC-V extension with
+  minimal adjustments.  We consider 32-bit CHERI-RISC-V less mature
+  and expect future disruptive modifications as it transitions to a more mature
   status.
 
 \item[Arm Morello] is an experimental instantiation created by Arm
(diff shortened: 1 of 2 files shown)


brooksdavis:
Seems reasonable to me.

arichardson (approved):

chap-architecture.tex
```diff
@@ -36,8 +36,9 @@ \section{Architectural Instantiations of CHERI Protection}
   implementations, adaptations of our CheriBSD and CheriFreeRTOS operating
   systems, Clang/LLVM/LLD toolchain, GDB debugger, and application suite.
 
-  We expect continuing disruptive modification to this ISA mapping, including
-  reencoding of many key instructions, as it transitions to a more mature
+  We aim to propose 64-bit CHERI-RISC-V as a RISC-V extension with
+  minimal adjustments.  We consider 32-bit CHERI-RISC-V less mature
+  and expect future disruptive modifications as it transitions to a more mature
```
rwatson, line 41: I’m good with committing this text, but it would be worth thinking about whether an example or two here would help explain why we think that -- e.g., that the capability encoding would ideally provide more bounds precision, and perhaps invest fewer bits in permissions.

rwatson (approved):

chap-architecture.tex
bsdjhb: I have added a small paragraph in the RISC-V chapter that tries to explain this a bit further including the specific examples you gave.

chap-cheri-riscv.tex
```diff
@@ -109,6 +109,14 @@ \subsection{Target RISC-V ISA Variants}
 We also describe extensions to RVS, the supervisor extension defined in the
 privileged portion of the ISA.
 
+We view 64-bit CHERI-RISC-V as a mature specification suitable for
+consideration as an official RISC-V extension.  However, we feel that
```
rwatson, line 113: I wonder if this is slightly strong, and we want to suggest that our current specification is suitable as a starting point or input for standardisation?

bsdjhb: Yes, this is a tad strong.  I do think we want the two paragraphs changed in this PR to be consistent.  I also think we want to not make it too weak either.

PR #21, opened 2022-10-31
Rename Permit_CInvoke to Permit_Invoke.

### LINKED ISSUE #20: PERMIT_CINVOKE should be PERMIT_INVOKE for consistency

bsdjhb:
This is more consistent with, for example, Permit_Seal gating use of CSeal.

Fixes #20

### DIFF
--- a/preamble.tex
+++ b/preamble.tex
@@ -380,7 +380,7 @@
 \makeatother
 \makecapperm{ASR}{Access\_System\_Registers}
 \makecapperm{Cid}{Set\_CID}
-\makecapperm{CInvoke}{CInvoke}
+\makecapperm{Invoke}{Invoke}
 \makecapperm{L}{Load}
 \makecapperm{LC}{Load\_Capability}
 \makecapperm{S}{Store}
@@ -391,6 +391,10 @@
 \makecapperm{X}{Execute}
 % No Permit_, so always use starred form even if not given
 \makecapperm*{G}{Global}
+\makeatletter
+% Legacy until Sail is updated for Perm_Invoke
+\ea\ea\ea\let\csname @capperm@\detokenize{Permit\_CInvoke}\ea\endcsname\csname @capperm@\detokenize{Permit\_Invoke}\endcsname
+\makeatother
 
 \makeatletter
 \newcommand{\@insnlabelname}[2]{insn:#1:#2}
--- a/app-cheri-128.tex
+++ b/app-cheri-128.tex
@@ -251,7 +251,7 @@ architectural bit\# & \cmuperms{} bit\#& Name \\
 \cperms{}[5] & 5 & \cappermSC \\
 \cperms{}[6] & 6 & \cappermSLC \\
 \cperms{}[7] & 7 & \cappermSeal \\
-\cperms{}[8] & 8 & \cappermCInvoke \\
+\cperms{}[8] & 8 & \cappermInvoke \\
 \cperms{}[9] & 9 & \cappermUnseal \\
 \cperms{}[10] & 10 & \cappermASR \\
 \cuperms{}[15--18] & 11--14 & Software-defined permissions \\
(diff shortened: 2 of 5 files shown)


jrtc27:
This will screw up the styling of Permit_CInvoke in the saildoc-generated descriptions; preamble-saildoc-macros has some magic to turn any `\textbf{Foo}` into `\@capperm@Foo` if such a macro exists, so if you rename the CInvoke macro that'll stop rewriting the CInvoke instances and leave them as bold rather than small caps.

bsdjhb:
Hmm, I checked the Sail for the CInvoke instruction and it seems to have both permit_cinvoke and permit_execute (the only permissions checked for that instruction) rendered the same.  All lower-case without any formatting.

bsdjhb:
(I do think though this does mean that the Sail model needs to change as well for the rename?)

jrtc27:
image
That's how the latest main GitHub Actions-built PDF looks

bsdjhb:
Ah, I looked at the actual Sail, not the exceptions prose below it.  I don't think I can easily remap @capperm@Permit\_CInvoke to Permit_Invoke text (and it wouldn't match the raw Sail), so I just defined both permissions for now.

jrtc27:
> Ah, I looked at the actual Sail, not the exceptions prose below it. I don't think I can easily remap @capperm@Permit_CInvoke to Permit_Invoke text (and it wouldn't match the raw Sail), so I just defined both permissions for now.

`\let\@capperm@Permit_CInvoke\@capperm@Permit_Invoke` (under `\makeatletter`) should do it

bsdjhb:
It seems to not like the second '@':

```
./preamble.tex:397: LaTeX Error: Missing \begin{document}.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
\GenericError  ...                                
                                                  \endgroup 
<to be read again> 
                   C
l.397 \let\@capperm@Permit_C
                            Invoke\@capperm@Permit_Invoke 
```

It is under `\makeatletter` in the source:

```
% No Permit_, so always use starred form even if not given
\makecapperm*{G}{Global}

\makeatletter
% Legacy until Sail is updated for Perm\_Invoke
\let\@capperm@Permit_CInvoke\@capperm@Permit_Invoke
```

I tried various incantations of using `\csname` around the names instead but those did not work either

jrtc27:
Bah underscores... maybe a better way to do it but this works:
```
diff --git a/preamble.tex b/preamble.tex
index 0eb1334..aca62de 100644
--- a/preamble.tex
+++ b/preamble.tex
@@ -381,8 +381,6 @@
 \makecapperm{ASR}{Access\_System\_Registers}
 \makecapperm{Cid}{Set\_CID}
 \makecapperm{Invoke}{Invoke}
-% Legacy until SAIL is updated for Perm\_Invoke
-\makecapperm{CInvoke}{CInvoke}
 \makecapperm{L}{Load}
 \makecapperm{LC}{Load\_Capability}
 \makecapperm{S}{Store}
@@ -393,6 +391,10 @@
 \makecapperm{X}{Execute}
 % No Permit_, so always use starred form even if not given
 \makecapperm*{G}{Global}
+\makeatletter
+% Legacy until SAIL is updated for Perm\_Invoke
+\ea\ea\ea\let\csname @capperm@\detokenize{Permit\_CInvoke}\ea\endcsname\csname @capperm@\detokenize{Permit\_Invoke}\endcsname
+\makeatother
 
 \makeatletter
 \newcommand{\@insnlabelname}[2]{insn:#1:#2}
```

preamble.tex
```diff
@@ -393,6 +393,9 @@
 \makecapperm*{G}{Global}
 
 \makeatletter
```
jrtc27, line 395: Eh, I deliberately kept it in a separate \makeatletter/\makeatother pair as it's unrelated to the commands that follow but I guess this is ok for something that *should* be short-lived

jrtc27 (approved):

preamble.tex
bsdjhb: Hmm, I could do that, doesn't hurt.

PR #22, opened 2022-11-01
Move CHERI enable bit to [ms]envcfg.

bsdjhb:
This was originally part of #10 but has been split out into its own PR.

The Xstateen extension did not exist when the spec was written originally.  This moves the enable bit to make use of that extension and slightly expands the definition of what this bit would mean.  In QEMU we don't fully implement the bit (it's hardwired to on).  QEMU also implements additional bits for load-side barriers in xCCSR, so we might want to defer removing the register outright (should we document those bits?).

### DIFF
--- a/chap-cheri-riscv.tex
+++ b/chap-cheri-riscv.tex
@@ -217,6 +217,8 @@ The following changes are specific to CHERI-RISC-V:
   as more specific cause information and the identity of the faulting
   register is reported in the existing \xtval{} CSRs (see
   Section~\ref{subsection:riscv:cheri-exception-reporting}).
+\item A new bit is defined in \menvcfg{} and \senvcfg{} to enable
+  CHERI support.
 \item New per-mode capability CSRs are added as \xccsr{} (see
   Section~\ref{subsubsec-ccsrs}).
 \item CHERI-related page permissions are added to RISC-V architectural
@@ -473,6 +475,28 @@ corresponding privilege mode's CSRs. Debate still open over whether these are en
 \label{tab:risc-v-access-system-registers-whitelist}
 \end{table}
 
+\subsubsection{CHERI Extension Control}
+
+A new bit in the \menvcfg{} and \senvcfg{} CSRs is used to enable
+CHERI for lower privilege levels.  When CHERI is disabled, attempting
+to execute CHERI-specific instructions will raise an illegal
+instruction fault including loads and stores which use a capability
+register (excluding the implicit \DDC{} operand for legacy
+loads/stores) as the memory operand.
+
+Other CHERI extensions are always enabled regardless of the state of
+this bit.  Specifically, bounds and permissions on \PCC{} and \DDC{}
+are always honored.  Exceptions always copy \PCC{} to \xEPCC{} on
+exception entry and restore the full \PCC{} on exception return.
+Capability mode is always honored if enabled in \PCC{}.  Software
+which disables CHERI in lower modes must take care to ensure that
+\PCC{} and \DDC{} are set to suitable values while lower modes
+execute.
+
+Bit 28 (\texttt{0x1C}) in the \menvcfg{} and \senvcfg{} CSRs is
+defined as the CHERI enable bit.  Its allocation within these CSRs may
+change until CHERI is ratified as a RISC-V extension.
+
 \subsubsection{Capability Control and Status Registers (CCSRs)}
 \label{subsubsec-ccsrs}
 New per HART \xccsr{} \texttt{XLEN}-bit RISC-V CSRs are defined as per
@@ -482,8 +506,7 @@ Figure~\ref{fig-ccsr} (shown for XLEN=32):
 \begin{center}
 \begin{bytefield}[bitwidth=\textwidth/34]{32}
   \bitheader[endianness=big]{0,1,2,31} \\
-  \bitbox{31}{\textbf{WIRI}}
-  \bitbox{1}{\texttt{e}}
+  \bitbox{32}{\textbf{WIRI}}
 \end{bytefield}
 \caption{\xccsr{} register format; WIRI bits are Write Ignore Read Ignore.}
 \label{fig-ccsr}
@@ -497,10 +520,6 @@ Figure~\ref{fig-ccsr} (shown for XLEN=32):
   The RISC-V spec unsatisfyingly does not specify what happens if
   software doesn't preserve values for WPRI, but then this spec
   doesn't need to either.}
-
-\begin{description}
-\item [e] The \texttt{e} ``enable'' bit \ajnote{read only? / only wriFable in u-mode? only writable with a specific permission in pcc?} tells whether capability extensions are enabled or disabled.
-\end{description}
 
 \subsection{Special Capability Registers (SCRs)}
 \label{subsection:cheri-riscv-scrs}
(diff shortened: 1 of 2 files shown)


chap-cheri-riscv.tex
```diff
@@ -502,34 +484,22 @@ \subsubsection{Capability Exception Reporting}
...
+\subsubsection{CHERI Extension Control}
 
-\pmnote{WIRI has been removed from the RISC-V specification.  The spec
-  now typically uses WPRI (Write-Preserve, Read-Ignore) for such
-  undefined bits (e.g. in \texttt{mstatus}).  Both WARL (Write-Any,
-  Read-Legal) and WLRL (Write-Legal, Read-Legal) seem unsuitable here.
-  The RISC-V spec unsatisfyingly does not specify what happens if
-  software doesn't preserve values for WPRI, but then this spec
-  doesn't need to either.}
+For processors supporting the \texttt{Smstateen} extension, bit 2 of
```
arichardson, line 489: I'm not quite clear on what the difference is between `menvcfg` (privspec 1.12) and [`mstateen`](https://github.com/riscv/riscv-state-enable/blob/main/content.adoc)? Both seem to be used to define access to extensions. I think we definitely need the `mstateen` to gate access to CHERI registers and SCRs, but do we also need a bit in menvcfg to limit access to the instructions?

arichardson: I guess the only instructions we have that don't access the registers are CRRL/CRAM and there shouldn't be much harm in always providing them (unless we want to have a mode where CHERI is  not available at all and all non-baseline ISA encodings trap).

bsdjhb: I was not aware of `menvcfg` when I wrote this text.  Certainly it might be that we might want that either in addition or possibly in place of just `mstateen`.  In particular, CHERI is not really just a side extension with some state like floating-point or crypto acceleration but instead makes non-trivial changes to how the base ISA is used.

jrtc27: In theory if you don't touch CHERI state you shouldn't know it's there, and that's true other than for memory accesses that wrap the address space where we will erroneously trap with a bounds fault even if DDC/PCC covers the whole address space

arichardson: I'm a bit concerned that bit 2 will conflict with the next extension that gets added to RISC-V. Should we pick something further the middle (e.g. 0xc/0x1c for CHERI?).

arichardson: So if they are added in order then bit 31 is safest for now since it won't conflict with anything until all bits are taken?

bsdjhb: I guess so.  We probably want some sort of note that the bit is subject to change when/if CHERI is ratified upstream.  We could perhaps use bit 0 (defined for "custom state") for now until CHERI is ratified?  I also need to go read the `senvcfg` spec to see how that applies here and if we want both bits or only one of them.

chap-cheri-riscv.tex
```diff
@@ -502,34 +484,22 @@ \subsubsection{Capability Exception Reporting}
...
+\mstateen{} and \sstateen{} is used to enable CHERI.  When CHERI is
+disabled, attempting to execute CHERI-specific instructions will raise
+an illegal instruction fault including loads and stores which use a
+capability register as the memory operand.
```
arichardson, line 493:
```suggestion
capability register (excluding the implicit \DDC{} operand for legacy loads/stores) as the memory operand.
```

arichardson (commented):
This loses the ability to disable CHERI for the current ring, but using a standardized mechanism sounds good to me.

We should probably use bit 0 until an official bit can be reserved?

bsdjhb:
The description of `senvcfg` in the current priv spec is not very clear.  It doesn't really define an overall purpose for the register, though it does seem perhaps a more apt place to define an enable bit for a "feature" vs the `XstateenX` which seem more for controlling access to banks of registers more like XSAVE extensions on x86.  `Xenvcfg` doesn't have a provision for local extensions like bit 0 though that can be used while waiting to be ratified which is rather inconvenient.

chap-cheri-riscv.tex
bsdjhb: I have read a bit more of some of the specs using `Xenvcfg` btw, and I do think that actually it is a better fit.  Given how this bit is currently defined, once CHERI is disabled you can't access the upper bits of registers that have been extended to capabilities or entirely new capability registers (such as DDC).  As such, I don't think it really makes sense to think of this in terms of enabling/disabling additional state.  Probably it is more about controlling the ability to execute new CHERI instructions for which `Xenvcfg` as used by the `Zcmo*` extensions is a better fit.  There is no reserved bit for local state, so we'd have to just pick a bit that fits in the lower 32 for now (probably bit 8 above `CBZE`) and just add a note that it is subject to change based on ratification.  Existing extension proposals include similar language about the bits possibly moving.

bsdjhb:
This version has a fixup commit at the end to move the enable bits to `*envcfg`.  If we want to go that route I would  fold it back into the first commit before merging.

chap-cheri-riscv.tex
```diff
@@ -473,34 +456,26 @@ \subsubsection{Controlling Access to CSRs}
...
+exception entry and restore the full \PCC{} on exception return.
+Capability mode is always honored if enabled in \PCC{}.  Software
+which disables CHERI in lower modes must take care to ensure that
+\PCC{} and \DDC{} are set to suitable values while lower modes
+execute.
 
-\begin{description}
-\item [e] The \texttt{e} ``enable'' bit \ajnote{read only? / only wriFable in u-mode? only writable with a specific permission in pcc?} tells whether capability extensions are enabled or disabled.
-\end{description}
+Bit 8 in the \menvcfg{} and \senvcfg{} CSRs is defined as the CHERI
```
arichardson, line 476: My only concern is that 8 will conflict with the next allocated bit -- maybe we should use something like 0x1c which is right in the middle so should be safe for a while?

chap-cheri-riscv.tex
```diff
@@ -473,34 +456,26 @@ \subsubsection{Controlling Access to CSRs}
...
-  \bitheader[endianness=big]{0,1,2,31} \\
-  \bitbox{31}{\textbf{WIRI}}
-  \bitbox{1}{\texttt{e}}
-\end{bytefield}
-\caption{\xccsr{} register format; WIRI bits are Write Ignore Read Ignore.}
-\label{fig-ccsr}
-\end{center}
-\end{figure}
+A new bit in the \menvcfg{} and \senvcfg{} CSRs is used to enable
+CHERI.  When CHERI is disabled, attempting to execute CHERI-specific
```
arichardson, line 462: "for lower privilege levels"?

arichardson (commented):
I am overall happy with this although I wonder how this will interact with future xCCR changes such as DDC/PCC interposition..

chap-cheri-riscv.tex
bsdjhb: Ok.

bsdjhb:
> I am overall happy with this although I wonder how this will interact with future xCCR changes such as DDC/PCC interposition..

Yeah, given the DDC/PCC offsetting PR, I might now defer removing xCCR for now and just make this PR be about moving the enable bit.

bsdjhb:
I've dropped the change to remove xCCSR for now.

arichardson (approved):
I think we might need a follow-up change to define the reset state here - to avoid lower rings abusing DDC to break CHERI-unaware SW I think it needs to be disabled on reset. But not necessarily as part of this PR

PR #23, opened 2023-01-03
Add CGetTop/CGetHigh/CSetHigh

arichardson:
Using the sail code from https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/55


nwf (approved):
LGTM

PR #24, opened 2023-01-04
X86 security extensions

jonwoodruff:
This is a new section to compare with recent x86 security extensions.  Some extensions may be missing, and bits of these sections may be missing as well.


bsdjhb (commented):
This generally seems fine to me.  I had thought that the ENDBRANCH part of CET was a subset of what sentries provide (such that you could dispense with that part of CET if desired).  In particular, IIRC CET has some somewhat convoluted ways to handle exceptions (branches that should still be allowed without using ENDBRANCH), and that the management of sentries gives an easier way to permit policy around which branches are protected.

Related to this last is that it may make sense (we should at least add a rationale note?) to have a way to require branches to use sentries, at least for user mode (e.g. via a bit in CR4 that supervisor mode could set).

bsdjhb:
Oh, it would probably be good to collapse the history of this branch down and use git rebase (instead of git merge) to simplify the history.

jrtc27:
Requiring branch targets to always be sentries would add a pointless instruction for jump tables; the current lowering for CHERI-RISC-V and Morello translated to CHERI-X86-64 (assuming code model with max 2 GiB code) would be something like:

```
foo:
.Lfoo$jump_table_base
  ...
  leac .LJTIm_n(%cip), %ctable
  movslq (%ctable,%rindex,4), %roffset
  leac .Lfoo$jump_table_base(%cip), %ctarget
  addc %ctarget, %roffset
  jmpc *%ctarget
```

Requiring sentries would put a `csealentry %ctarget` in between the last two instruction which just bloats things.

jrtc27:
Although doing `leac	.LJTIm_n(%cip), %ctable` rather than `movc	.LJTIm_n@GOTPCREL(%cip), %ctable` is a bit dodgy actually, if there are cases the compiler believes are impossible but end up happening, which one can easily construct... then you get an untrapped OOB read that is then used as an offset to jump to. Should fix CHERI-RISC-V and Morello to be more conservative...

brooksdavis:
This PR would benefit from rebasing.

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
 
 If the \insnxesref{XCHG} instructions \texttt{91} - \texttt{97} are not
 commonly used, they could be removed in capability mode.
+
+\section{Relationship to x86 Security Extensions}
```
bsdjhb, line 1264: Add "Existing" before x86 perhaps?

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+of read/write permissions of 15 sets of pages.
+The page table leaf entries themselves have a new 4-bit field to
+identify a page set (or key), with the value 0 wired to unfiltered.
+MPK was introduced with Skylake in 2015, and only appears to have
+been enabled for Xeon processors.
+AMD's Zen3 processors also appear to have MPK support since 2020,
+presumably enabled for both consumer and professional variants.
+
+In userspace, a thread may write its PKRU register, which maps
+16 pairs of read/write permissions coorosponding to the 16 possible
```
bsdjhb, line 1398:
```suggestion
16 pairs of read/write permissions corresponding to the 16 possible
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+keys.
+By writing this register, a thread may quickly remove access to subsets
+of pages.
+Current implementations appear to flush the back end of the pipeline,
+incuring a cycle penalty of roughly 20 cycles.
```
bsdjhb, line 1403:
```suggestion
incurring a cycle penalty of roughly 20 cycles.
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+This filtering is naturally per-thread, unlike the page table iteself
```
bsdjhb, line 1404:
```suggestion
This filtering is naturally per-thread, unlike the page table itself
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+access for a few sensitive structures.
+For example, secret keys might be stored in a region of memory that
+is only briefly made accessible when necessary.
+
+MPK has 3 major disadvantages compared to CHERI capabilities:
+scaling, PKRU protection, and buffer sharing.
+
+While MPK appears to be extremely unscalable on the surface, work
+has been done to virtualise the page sets with an API similar to
+memprotect so that sets of pages can have their access permissions
```
bsdjhb, line 1415:
```suggestion
mprotect so that sets of pages can have their access permissions
```

bsdjhb: Not sure if you intend a specific memprotect() function distinct from the POSIX mprotect()?  If so, this might need a citation?

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+modified arbitrarily.
+This can actually be implemented in the backend with mprotect
+(with process-global semantics), or can be implemented with MPK
+for sets whose permissions are changed frequently.
+Nevertheless, there are likely to be serious scaling problems
+if the abstraction library needs to resort to system calls that
+traverse page tables to adjust permissions or mask ids.
+
+PKRU register is freely writable in user-mode.
```
bsdjhb, line 1424:
```suggestion
The PKRU register is freely writable in user-mode.
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+
+MPX introduced four new 128-bit BND registers to hold both an
+upper and a lower bound for a range, along with instructions
+to move bounds between registers and memory.
+MPX also introduced instructions to create bounds and compare integer
+values to the bounds;
+upper and lower bounds are compared with separate
+instructions.
+
+MPX further introduces two table access instructions
```
bsdjhb, line 1459:
```suggestion
MPX further introduced two table access instructions
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+for loading and storing bounds for any pointer in memory
+from a table.
+Each leaf entry in the table is 32-bytes, including 8 bytes for each
```
bsdjhb, line 1462:
```suggestion
Each leaf entry in the table was 32-bytes, including 8 bytes for each
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+of the bounds and 8-bytes to hold a copy of the expected pointer value.
+Using these instructions, a program can be compiled to stash
+bounds for each pointer in memory in a shadow space, loading
+and storing bounds whenever pointers themselves are loaded and stored.
```
bsdjhb, line 1466:
```suggestion
Using these instructions, a program could be compiled to stash
bounds for each pointer in memory in a shadow space, loading
and storing bounds whenever pointers themselves were loaded and stored.
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+\subsubsection{MPX Shortcomings}
+
+MPX had critical shortcomings from two directions: security and
+performance.
+
+MPX was designed to maximise compatibility.
+As a result, MPX "failed open" in many cases where unexpected
+cases arose.
+If the pointer from the bounds table did not match the pointer
+from memory, bounds were left in state that would not fault.
```
bsdjhb, line 1477:
```suggestion
from memory, bounds were left in a state that would not fault.
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+This causes the hardware to allocate a number of structures in a
+reserved region of DRAM to track memory for the enclave before
+executing the next instruction.
+The EADD instruction copies a page of common memory into a
+newly-allocated enclave page, keeping a running hash of its contents.
+
+CHERI relies on simple invariants that can be enforced within
+single-cycle instructions.
+Intel SGX instead trusts the correctness and atomicity of
+hyper-visor-like subroutine calls due to their being
```
bsdjhb, line 1542:
```suggestion
hypervisor-like subroutine calls due to their being
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+provided by processor firmware.
+
+On one hand, the SGX precedent suggests that trusted hyper-visor-like
```
bsdjhb, line 1545:
```suggestion
On one hand, the SGX precedent suggests that trusted hypervisor-like
```

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+
+In addition, CHERI-x86-64 implementors should resist the pressure
+to leave CHERI de-optimised under the assumption that security
+is a niche application.
+CHERI is designed for overheads within 5\% of unprotected
+performance in most cases, but a few corner cases in forwarding
+or branch prediction can easily push overheads over 20\% and
+greatly restrict adoption.
+
+\subsection{Intel SGX (deprecated)}
```
bsdjhb, line 1523: I don't think SGX is deprecated fully (unlike MPX).  Intel has dropped it from "consumer" CPUs, but Xeons still ship it I believe according to https://community.intel.com/t5/Blogs/Products-and-Solutions/Security/Rising-to-the-Challenge-Data-Security-with-Intel-Confidential/post/1353141

chap-cheri-x86-64.tex
```diff
@@ -1260,3 +1260,300 @@ \subsection{XCHG [ER]AX Opcodes}
...
+
+CHERI relies on simple invariants that can be enforced within
+single-cycle instructions.
+Intel SGX instead trusts the correctness and atomicity of
+hyper-visor-like subroutine calls due to their being
+provided by processor firmware.
+
+On one hand, the SGX precedent suggests that trusted hyper-visor-like
+subroutines can be implemented in the CPU.
+On the other hand, Intel SGX was only implemented in a single Intel
```
bsdjhb, line 1547: I'm not sure these next two sentences are accurate.  Intel's current top-of-line Xeon processors ship SGX: https://www.intel.com/content/www/us/en/products/details/processors/xeon/scalable/platinum.html

PR #25, opened 2023-01-12
Update CHERI Concentrate compression listings

arichardson:
The listings now correspond to https://github.com/CTSRD-CHERI/cheri-cap-lib/commit/7d9fdce16a8ed8def1b5bd3f73109876179d6479

### DIFF
--- a/app-cheri-concentrate-listings.tex
+++ b/app-cheri-concentrate-listings.tex
@@ -4,6 +4,8 @@
 The cheri-cap-lib~\cite{CHERI-cheri-cap-lib} open-source library has been used in all of our open-source~\cite{CHERI-cheri-cpu,CHERI-Piccolo,CHERI-Flute,CHERI-Toooba} implementations and contains certain notable algorithms which have been highly optimised and verified to a significant degree.
 These algorithms are well documented by their contents, so several are reproduced in their current form here to serve as reference to anyone implementing CHERI Concentrate compression in any model or microarchitecture.
 
+% The current listings here correspond to https://github.com/CTSRD-CHERI/cheri-cap-lib/commit/7d9fdce16a8ed8def1b5bd3f73109876179d6479
+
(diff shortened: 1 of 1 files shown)


PeterRugg:
Hmm, this includes my changes for the single cycle in-bounds check on CSetBounds, which is perhaps on the unstable side to include in the arch doc at the moment (evaluating it more thoroughly currently).
Also, do we think manual copies like this are the appropriate way forward, or would we want to pull it in automatically (or as a submodule) from cheri-cap-lib, as we do for Sail?

arichardson:
> Hmm, this includes my changes for the single cycle in-bounds check on CSetBounds, which is perhaps on the unstable side to include in the arch doc at the moment (evaluating it more thoroughly currently). Also, do we think manual copies like this are the appropriate way forward, or would we want to pull it in automatically (or as a submodule) from cheri-cap-lib, as we do for Sail?

Oh good point, can you suggest a "stable" commit hash for this sync?

Longer-term, I agree that we should include the code instead of copying since it will get out-of-sync.

PeterRugg:
Hmm, just before this one, so https://github.com/CTSRD-CHERI/cheri-cap-lib/commit/be80fbe51c0860394807980cf1823d229aeeaca7 is probably a good bet. Hopefully I'll be confident in the new check soon. I'm happy to have a shot at setting it up as per sail unless there's a reason to put off doing that?

arichardson:
> Hmm, just before this one, so [CTSRD-CHERI/cheri-cap-lib@be80fbe](https://github.com/CTSRD-CHERI/cheri-cap-lib/commit/be80fbe51c0860394807980cf1823d229aeeaca7) is probably a good bet. Hopefully I'll be confident in the new check soon. I'm happy to have a shot at setting it up as per sail unless there's a reason to put off doing that?

Sure, I'll close this PR then.

PR #26, opened 2023-01-25
Automate extraction of concentrate listings

PeterRugg:


arichardson (approved):
Thanks, that looks good to me. What commit of the cap-lib do those changes correspond to?

I don't think it's necessary for this change but we could also generate a cap-lib-version.tex that contains a link to cap-lib at the version used to generate those listings.

PeterRugg:
@arichardson Good idea! I've done that. I note that we don't have any URLs inline in the document otherwise, so I hope this isn't too distasteful. I think it is helpful though.

PeterRugg:
Oh, hmm, the text refers to line numbers... It also looks like `GetTop` was originally intended to be below `GetBase`, which isn't there at all now? I should rework it before I merge.

PR #27, opened 2023-01-27
Make the CHERI exception codes architecture-specific.

bsdjhb:
Move the tables out of the architecture chapter and into each architecture spec.  Morello already uses a different scheme for reporting exceptions that doesn't follow this code.

For RISC-V I moved the description of xtval out into a new subsection next to exception handling rather than being buried in a section about new CSRs.

For both RISC-V and x86-64 I have kept the existing values, but have condensed the tables slightly since they otherwise float several pages away.

### DIFF
--- a/chap-cheri-riscv.tex
+++ b/chap-cheri-riscv.tex
@@ -216,7 +216,7 @@ The following changes are specific to CHERI-RISC-V:
   Additional capability-specific exception cause information, such
   as more specific cause information and the identity of the faulting
   register is reported in the existing \xtval{} CSRs (see
-  Section~\ref{subsubsec-cheri-tval}).
+  Section~\ref{subsection:riscv:cheri-exception-reporting}).
 \item New per-mode capability CSRs are added as \xccsr{} (see
   Section~\ref{subsubsec-ccsrs}).
 \item CHERI-related page permissions are added to RISC-V architectural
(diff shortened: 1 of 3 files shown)


bsdjhb:
Note that unaligned base is not listed in the exception priority for RISC-V.

One other thought is that Morello uses far fewer exception codes.  It has a single "Permission" fault that reports any permission fault rather than separate codes per permission.  Morello is also missing the "Type" violation fault it seems.

For x86 I'd like to compress the values down to reclaim some of the gaps, but I'm also considering if a single "Permission" code is sufficient in practice (the `si_code` values we report to userland just use a single code for all permission faults).

chap-cheri-riscv.tex
```diff
@@ -1008,6 +979,109 @@ \subsubsection{Exceptions to User Mode}
...
+general-purpose capability register file.
+\end{description}
+
+\begin{table}
+\begin{center}
+\begin{tabular}{ll}
+\toprule
+Value & Description \\
+\midrule
+0x00 & None \\
```
arichardson, line 1018: Would be nice if we could generate this table from sail, but almost certainly not worth the effort.

arichardson (approved):

PeterRugg (approved):

PeterRugg:
All looks good. Happy to add the unaligned_base priority once this is merged.

PR #28, opened 2023-01-30
Add more prose on direct memory-offset MOVs.

bsdjhb:
Recommend deprecating these instructions in capability mode and not extending the immediate memory offsets to support capability immediates.

### DIFF
--- a/chap-cheri-x86-64.tex
+++ b/chap-cheri-x86-64.tex
@@ -622,6 +622,17 @@ add %csp,$16
     \insnnoref{CMPXCHG} should default to integer operands in
     capability mode.
 
+  \item \insnnoref{XCHGC} will also be required to support atomic
+    operations on capabilities.
+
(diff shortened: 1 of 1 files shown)


chap-cheri-x86-64.tex
```diff
@@ -373,6 +374,17 @@ \subsubsection{Absolute Addresses}
 register.  These operands are always evaluated as offsets relative to
 \DDC{} including in capability mode.
 
+\subsubsection{Direct Memory-Offset MOVs}
+
+The direct memory-offset \insnnoref{MOV} instructions store the
+absolute address of a memory operand as an immediate operand.
+Extending these instructions to support capability immediates would
+require padding nops to align the capability immediate.  However, we do
```
jrtc27, line 382: as well as text relocations, even for position-dependent code, which is the bigger nope

PR #29, opened 2023-02-21
Fix trailing whitespace and add a pre-commit hook

arichardson:
To enable the hook, use `pip install --user pre-commit && pre-commit install`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,14 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+exclude: "^sail_latex_riscv/.*"
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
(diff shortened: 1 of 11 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
+    -   id: mixed-line-ending
+        args: [ --fix=lf ]
+    - id: check-case-conflict
```
jrtc27, line 14: Some weird spacing in this file

jrtc27 (commented):
Seems it would be worth running it in CI for people who haven't set it up locally?

bsdjhb (approved):
Seems ok to me.  Presumably you will squash the history a bit (at least squashing the last two together?)

PR #30, opened 2023-03-07
x86: Renumber capability fault error codes.

bsdjhb:
- Remove the unused 'None' value.

- Renumber tag violation first since to better match exception priority.

- Collapse error codes to remove gaps.

### DIFF
--- a/chap-cheri-x86-64.tex
+++ b/chap-cheri-x86-64.tex
@@ -1068,27 +1068,21 @@ the specific violation.
 \toprule
 Value & Description \\
 \midrule
-0x00 & None \\
-0x01 & Length Violation \\
-0x02 & Tag Violation \\
-0x03 & Seal Violation \\
-0x04 & Type Violation \\
-0x05-0x07 & \emph{reserved} \\
-0x08 & Software-defined Permission Violation \\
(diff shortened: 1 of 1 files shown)


bsdjhb:
Note that Morello uses a single exception for all permission violations rather than identifying specific permission violations, so one option might be to do something similar here if the granularity is not really needed.

jonwoodruff (approved):
Seems fine! I guess there's no function for "None" since we have indicated that there was a capability-type exception somewhere else?

bsdjhb:
On x86 exceptions invoke one of 256 function pointers from a table (the first 32 function pointers are for exceptions, the remaining ones are for external interrupts from PICs).  The way an OS distinguishes exceptions is to install unique handlers for the various function pointers.  Capability exceptions would use a dedicated slot in the first 32 function pointers, so the code handling the function would already know it's a capability exception.  The error code is a value pushed onto the stack by the processor to provide further details about the exception.  The error code discussed here is the value that gets pushed onto the stack.

One open question is if we really need the broken out values for individual permission bits, or if we want to have a single "permission violation" bit which is kind of how it works on Morello.  On Morello there are just 4 distinct capability exceptions.

PR #31, opened 2023-03-09
X86 lea

bsdjhb:


jonwoodruff (approved):
Seems sensible.  The new note is a bit wordy for a spec, but is certainly useful justification for people who might provide feedback on the spec.

PR #32, opened 2023-03-09
x86: Mark other status flags in RFLAGS undefined.

bsdjhb:
If a CHERI instruction modifies any of the status flags in RFLAGS, claim that the other status flags are undefined so that the values do not have to be preserved.

### DIFF
--- a/chap-cheri-x86-64.tex
+++ b/chap-cheri-x86-64.tex
@@ -622,6 +622,15 @@ add %csp,$16
     \insnnoref{CMPXCHG} should default to integer operands in
     capability mode.
 
+  \item \insnnoref{CMPXCHG2C} will be required to support atomic
+    operations on pairs of capabilities.
+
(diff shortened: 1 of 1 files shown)


bsdjhb:
This is another way of marking other status flags in RFLAGS as undefined using somewhat similar language to that used in the SDM for other instructions.  I believe it is also important to be explicit about which flags are set for new instructions (vs extensions of existing instructions like ADD where one is defining how the existing status flag semantics work in the extended form).

jonwoodruff (approved):
Looks like the right decisions here, provided the preponderance of precedent is toward leaving unrelated flags undefined.

bsdjhb:
Recently-added instructions such as `ADCX` or `BEXTR` use this approach of marking status flags undefined if they are not set to an explicit value (and if the instruction sets at least one status flag).

PR #34, opened 2023-03-23
Make DDC/PCC offsetting an optional feature rather than mandatory

bsdjhb:
This almost certainly needs some rewording.  The first commit was my first stab at just trying to deprecate the instructions, but I think the prose there is rather thin and probably the last commit is a better starting point.  Probably the x86 chapter change should be pulled out of this and handled separately once we have consensus on this more general change.

### DIFF
--- a/chap-rationale.tex
+++ b/chap-rationale.tex
@@ -616,11 +616,6 @@ have been with the \ctag{} bit set, and zero in the \cbase{} field and
 \clength{} fields.  Effectively, NULL would have been a capability for
 an array of length zero.
 
-Many CHERI instructions are agnostic as to which of these two
-conventions for NULL is employed, but the \insnref{CFromPtr} and
-\insnref{CToPtr} operations are aware of the
-convention.
-
 The advantages of NULL's \ctag{} bit being unset are:
 
 \begin{itemize}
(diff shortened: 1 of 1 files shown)


chap-rationale.tex
```diff
@@ -1465,4 +1460,84 @@ \section{Attempted Montonicity Violations Clear Tags}
...
+  the hybrid code be relocatable and position independent.  For simple
+  support of legacy 64-bit processes for which \DDC{} and \PCC{}
+  bounds covered the entire user portion of the address space with a
+  base address of 0 this did not matter.  However, this was a hurdle
+  for hybrid operating system kernels which tended to run in a higher
+  range of virtual addresses and were not always relocatable.  In
+  practice hybrid kernels ran with \DDC{} and \PCC{} whose bounds
+  spanned the entire address space.
+
+  \item The Morello architecture shipped with knobs to toggle the
```
jrtc27, line 1528: If this is meant to be nested then you need a nested itemize, otherwise this should be deindented

chap-rationale.tex
```diff
@@ -1465,4 +1460,84 @@ \section{Attempted Montonicity Violations Clear Tags}
...
+    disabled, \DDC{} and \PCC{} still constrained legacy memory
+    accesses via bounds and permissions, but legacy integer pointers
+    were interpreted as addresses rather than offsets.
+\end{itemize}
+
+CHERI no longer mandates \DDC{} and \PCC{} offsetting by default.
+CHERI architectures may provide it as an optional feature which can be
+enabled at runtime or may omit it entirely.  CHERI compilers always
+treat integer pointers as addresses using \insnref{CGetAddr} and
+\insnref[CSetAddr} to handle conversions between capabilities and
```
jrtc27, line 1539:
```suggestion
\insnref{CSetAddr} to handle conversions between capabilities and
```

arichardson (commented):
Thanks, I like this change and IMO we should make it optional for RISC-V as I don't believe mandatory offsetting should be included in the CHERI-RISC-V-MVP.

davidchisnall (approved):
Looks good to me. FWIW, we don't have a DDC at all in CHERIoT, since we don't aim to support code compiled with a non-CHERI-aware compiler (we can do a hybrid ABI by explicitly offsetting from a reserved register).

It might be worth mentioning the other compromise that's need discussed: DDC offsetting but with strong alignment requirements on DDC. For WASM-like use cases, requiring the top and bottom of DDC to be 1 MiB aligned would be fine and would significantly reduce the critical path length for the checks.

arichardson (approved):

PR #36, opened 2023-03-28
X86 opcodes

bsdjhb:
This is more of a draft for now.  I started creating more detailed opcode descriptions to add extra details I've been depending on in the toolchain and Sail modeling work.  However, after getting started down this path I think I probably need to move the instruction descriptions out into either a separate top-level section or even a separate chapter as we do for RISC-V (and did for MIPS).  I should also add similar details for existing opcodes extended by the 06 prefix (e.g. ADD).

There's a lot of repetition though so I probably need to gain some inspiration from @jrtc27's RISC-V instruction macros.

### DIFF
--- a/insn-x86-64/add.tex
+++ b/insn-x86-64/add.tex
@@ -1,0 +1,50 @@
+\clearpage
+\phantomsection
+\addcontentsline{toc}{subsection}{ADD - Add}
+\insnxeslabel{add}
+\subsection*{ADD - Add}
+
+\begin{x86opcodetable}
+  \xopcode{CAP + 05 \emph{id}}{ADD CAX\emph{, imm32}}
+  {I}{Valid}{Valid}
+  {Add \emph{imm32 sign-extended to 64-bits} to the address field of
+    CAX.}
+  \xopcode{CAP + 81 /0 \emph{id}}{ADD \emph{r/mc, imm32}}
+  {MI}{Valid}{Valid}
+  {Add \emph{imm32 sign-extended to 64-bits} to the address field of
+    \emph{r/mc}.}
+  \xopcode{CAP + 83 /0 \emph{ib}}{ADD \emph{r/mc, imm8}}
+  {MI}{Valid}{Valid}
+  {Add \emph{sign}-extended \emph{imm8} to the address field of
+    \emph{r/mc}.}
+  \xopcode{CAP + 01 \emph{/r}}{ADD \emph{r/mc, r64}}
+  {MR}{Valid}{Valid}
+  {Add \emph{r64} to the address field of \emph{r/mc}.}
+  \xopcode{CAP + 03 \emph{/r}}{ADD \emph{rc, r/m64}}
+  {RM}{Valid}{Valid}
+  {Add \emph{r/m64} to the address field of \emph{rc}.}
+\end{x86opcodetable}
+
+\begin{x86opentable}
+  \xopen{RM}{ModRM:Reg (r,w)}{ModRM:r/m (r)}{NA}{NA}
+  \xopen{MR}{ModRM:r/m (r,w)}{ModRM:reg (r)}{NA}{NA}
+  \xopen{MI}{ModRM:r/m (r,w)}{imm8/32}{NA}{NA}
+  \xopen{I}{CAX}{imm32}{NA}{NA}
+\end{x86opentable}
+
+\subsubsection*{Description}
+
+Adds the source operand to the \textbf{address} field of the
+destination operand and then stores the result in the destination
+operand. The destination operand can be a register or a memory
+location. The source operand can be an immediate, a register, or a
+memory location.
+
+If the new value of the \textbf{address} field makes the resulting
+capability unrepresentable, the \textbf{tag} field in the resulting
+capability is cleared.
+
+\subsubsection*{Flags Affected}
+
+The OF, SF, ZF, AF, CF, and PF flags are set according to the value of
+the resulting \textbf{address} field.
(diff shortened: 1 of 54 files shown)


bsdjhb:
This now covers all of the new and extended instructions in the current sketch.

insn-x86-64/add.tex
jonwoodruff, line 1: I'm confused about the use of "to the address" in this instruction.
ADD CAX, imm32
I thought this one would simply "IncOffset" CAX.
The explanation says "Add imm32 sign-extend to 64-bits to the address of CAX".
Does CAX have an address?
Maybe it means the address in CAX? As in, to the capability address in CAX?  I guess CAX is by definition a capability?

I think I've typed myself into knowing what you mean, but there should be a less confusing phrasing.
How about "to the address field of CAX" and so on?

bsdjhb: Yes, the prose paragraph below says address field, so using that makes sense.  I've fixed that here and in other instruction descriptions.

insn-x86-64/sub.tex
jonwoodruff, line 1: This one should also probably mention the representability check.

insn-x86-64/xadd.tex
jonwoodruff, line 1: More of a comment on the ISA; adding addresses is pretty weird.  I guess the common case here is to XADD with an integer in "r/mc"?

This one might mention representability as well.

insn-x86-64/xor.tex
jonwoodruff, line 1: This one should mention representability.

insn-x86-64/buildcap.tex
jonwoodruff, line 1: The Op/En column isn't wide enough on my build.

Is there a compelling reason why this one is not a 2-operand instruction? Presumably with the "bag of bits" being tagged destination?

insn-x86-64/lea.tex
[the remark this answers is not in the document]
bsdjhb: Yep, thanks.

insn-x86-64/xadd.tex
bsdjhb: Yeah, really this should perhaps be written as a 3-operand instruction even though only 2 operands are encoded similar to the alternate syntax we've considered for ADD.  The problem is that the this operand is treated as a r/m64 for the source operand, but r/mc for the destination operand.  I could make that apparent via the Op/En table by reusing the same modr/m field with the different types.  I might take a stab at that.

insn-x86-64/or.tex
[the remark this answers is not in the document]
bsdjhb: Ok, I've added a note to this effect for ADD, SUB, AND, OR, XOR, XADD, SCADDR, and SCOFF.

insn-x86-64/buildcap.tex
bsdjhb: I think for the case where you are running through multiple possible roots (the use case using CBuildCap + CCopyType + CCSeal to re-derive for swap-in for example) you want to be able to try to build the capability without hosing the destination in case something fails and you need to try the next root?

bsdjhb: Fixed the Op/En column width.

insn-x86-64/cram.tex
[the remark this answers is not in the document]
bsdjhb: I could say "aligned bounds".  That feels a bit redundant as CHERI bounds are always aligned, but that is also the point of this instruction.

insn-x86-64/xadd.tex
bsdjhb: I've pushed a test commit that does the 3 operand approach.  If you prefer it I will merge it into the original XADD commit.

insn-x86-64/buildcap.tex
jonwoodruff: When you fail, don't you just leave the tag alone? I thought the idea of that chain of instructions was that the capability would come out tagged at the end if any succeeded, so you would need the bag-of-bits input operand to be the output operand through the whole pipeline...

insn-x86-64/xadd.tex
jonwoodruff: I've managed to see the new one now; it is certainly more clear, if a bit complicated now.  Those kind of operand restrictions are not uncommon in x86 specs I think though?

insn-x86-64/cram.tex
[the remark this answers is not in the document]
jonwoodruff: New text is certainly more clear, if a bit verbose.

insn-x86-64/buildcap.tex
bsdjhb: Humm, I wonder if that's obvious to the compiler such that if you write C code like:

```
void
restore_tag(uintcap_t in, uintcap_t *out)
{
    uintcap_t cap, sealcap;
    u_int i;

    for (i = 0; i < NUM_ROOTS; i++) {
        cap = cheri_buildcap(roots[i], in);
        sealcap = cheri_copytype(roots[i], in);
        cap = cheri_condseal(cap, sealcap);
        if (cheri_gettag(cap)) {
            *out = cap;
            return;
        }
    }

    /* Failed to tag. */
    *out = in;
}
```

With 2 operands this will keep destroying 'in' requiring you to effectively do a MOV of `in` into `cap` before doing BUILD I think, especially as you need it as in input to cheri_copytype anyway.  Even without the loop just using the existing code in cheri_restore_tag in the swap pager you have the issue that you need the original input to pass to cheri_copytype.  Perhaps we could invoke cheri_copytype before cheri_buildcap to help with that, but still doesn't help with the loop case above and if the compiler can really figure out that in the case it's not tagged the final value of cap is the same as the original value of in.

jonwoodruff: As you say, putting copytype above buildcap means that buildcap and condseal can be naturally 2-operand.  Do we think that this is more sensitive to general-purpose performance than, say "or" or "and"?  It seems odd that x86 would have 2-operand instructions for such common cases, and we would go for 3-operand in esoteric cases like this.

Would using a MOV and a some 2-operand instructions end up being smaller in memory than 3-operand instructions by any chance?  Instruction fusion might then kick in so that the number of cycles would be a wash in most cases.

PR #38, opened 2023-04-14
Use ubuntu-latest for GitHub CI actions.

bsdjhb:
ubuntu-18.04 is no longer available and switching to latest avoids having to update this in the future.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


arichardson (commented):
Seems fine to me. I guess it could break in other ways (packages renamed, etc) but it's an improvement. How do we deal with all the duplication in the other repos? Does someone need to figure out how to write a shared action that we can use everywhere?

bsdjhb:
Humm, a shared action certainly would be nice

PR #39, opened 2023-04-18
Remove offsetting from CHERI-x86

bsdjhb:
This assumes (and implicitly depends on) #34.

### DIFF
--- a/chap-cheri-x86-64.tex
+++ b/chap-cheri-x86-64.tex
@@ -110,9 +110,8 @@ architectures:
 \begin{itemize}
 \item Tagged memory with capability-width tag granularity and alignment.
 \item Registers able to hold capabilities are tagged.
-\item \CIP{} transforms and controls program-counter-relative fetches.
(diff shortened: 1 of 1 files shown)


chap-cheri-x86-64.tex
```diff
@@ -134,6 +133,10 @@ \subsection{Unique Architectural Features}
  \item Requests for most non-monotonic capability transformations
    clear the tag on the destination capability rather than raising an
    exception.
+ \item \RIP{} is the full integer value (virtual address) of \CIP{}
```
bsdjhb, line 136: After #34 lands these will no longer be unique features.

bsdjhb: I will leave them in for now though since they are still unique until RISC-V is updated to handle offsetting differently.

jonwoodruff (approved):
I've read through all the changes and did not spot anything undesirable.
In the couple of awkward spots, I couldn't quickly think of anything better, and it wasn't incorrect.

PR #43, opened 2023-05-04
x86: Remove far branches from capability mode.

bsdjhb:
They are not needed to switch into 64-bit mode from capability mode and invalidating them removes the need to handle the awkward case of a far jump to a 32-bit segment from capability mode.

Note that IRET for exception return can still do a type of far branch.

### DIFF
--- a/chap-isaref-x86-64.tex
+++ b/chap-isaref-x86-64.tex
@@ -241,6 +241,17 @@ or 64).
   \hline
   A3 & MOV \emph{moffs8},[ER]AX & Move [ER]AX to  (\emph{offset}).\\
   \hline
+  CA & RET \emph{imm16} &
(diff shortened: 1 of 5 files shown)


chap-isaref-x86-64.tex
```diff
@@ -239,6 +239,17 @@ \section{Instructions Deprecated in Capability Mode}
   \hline
   A3 & MOV \emph{moffs8},[ER]AX & Move [ER]AX to  (\emph{offset}).\\
   \hline
+  CA & RET \emph{imm16} &
```
jrtc27, line 242: Hm, deprecated isn't quite the right term for all these, that implies they still exist but aren't recommended

bsdjhb: Actually, I fixed this more generally in #44.

PR #46, opened 2023-05-27
x86: Add [GS]CHI and GCLIM.

bsdjhb:


arichardson (approved):

insn-x86-64/schi.tex
```diff
@@ -0,0 +1,27 @@
...
+\begin{x86opentable}
+  \xopen{MR}{ModRM:r/m (r, w)}{ModRM:reg (r)}{NA}{NA}
+\end{x86opentable}
+
+\subsubsection*{Description}
+
+Sets the \textbf{high half} of the destination operand to the source
+operand and stores the result in the destination operand.  The
+destination operand can be a register or memory location; the source
+operand can be a register.  The \textbf{tag} field in the destination
```
arichardson, line 22: Looks like we already have a macro for this
```suggestion
operand can be a register.  The \ctag{} field in the destination
```

PR #50, opened 2023-06-08
x86: Don't set ZF for various GC* instructions.

bsdjhb:
There are probably not any hot paths that would need to optimize checking the result against zero.


bsdjhb:
I think originally I had thought "might as well always set ZF when fetching fields", but I suspect there's a uarch cost to setting the flags as well as ISA implications (e.g. it's sometimes useful that certain instructions don't alter RFLAGS so you can insert them between an instruction which modifies RFLAGS and a conditional branch).  I don't think it is critical to set RFLAGS for these instructions.

PR #52, opened 2023-06-08
x86 Capability Flags instructions

bsdjhb:


bsdjhb:
After talking a bit more with @jrtc27 and @arichardson today, I'm not convinced X86 needs a mode toggle instruction like BX on Morello.  Most of the use cases on Morello are to deal with addressing modes and X86 already solves that by using a single-byte prefix to enable alternate mode addressing on individual instructions.

PR #53, opened 2023-06-20
RISC-V: Remove DDC/PCC offsetting entirely for now.

bsdjhb:
This does not include any way to conditionally enable it though that could be added back in the future.  Mostly this is a draft to prompt further discussion.

Some interesting things: consistent with removing offsetting, this changes the description of SCRs that extend CSRs such that the CSR now reads/writes the address of the SCR rather than the offset.  Secondly, the discussion about unaligned PCC bases now very well all be moot without PCC offsetting (if PC is the address of PCC rather than the offset).  I've merely left a note about that rather than removing the section however.

### DIFF
--- a/chap-cheri-riscv.tex
+++ b/chap-cheri-riscv.tex
@@ -171,9 +171,8 @@ CHERI-RISC-V shares the following features with other CHERI architectures:
 \begin{itemize}
 \item Tagged memory with capability-width tag granularity and alignment.
 \item Registers able to hold capabilities are tagged.
-\item \PCC{} transforms and controls program-counter-relative fetches.
-\item \DDC{} transforms and controls legacy RISC-V load-store instructions,
-  including relocating access addresses using the capability base and offset.
+\item \PCC{} controls program-counter-relative fetches.
+\item \DDC{} controls legacy RISC-V load-store instructions.
 \item Floating point is fully supported, including capability-relative
   floating-point load and store instructions.
 \item General-purpose registers are extended to hold capabilities.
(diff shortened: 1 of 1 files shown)


bsdjhb:
NB: This does not remove or deprecated CFromPtr/CToPtr.

arichardson:
I've just uploaded a sail PR to make PCC relocation optional (which also removes all possibilities of generating the unaligned base exception): https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/77

bsdjhb:
I think this is now a non-draft candidate.  It doesn't solve how to permit optional relocation in U mode but leaves that for a future version.

chap-cheri-riscv.tex
```diff
@@ -1560,3 +1503,16 @@ \subsection{Separate Capability Register File}
...
+CHERI-RISC-V originally specified that legacy memory accesses using
+integer pointers were relocated by \DDC{} and \PCC{} in addition to
+being constrained.  In this model, integer pointers were treated as
+offsets relative to the base of \DDC{} and \PCC{} rather than
+addresses.
+
+The current version of CHERI-RISC-V does not relocate integer
+pointers.  However, it may be desirable to provide optional relocation
+in the future, in particular to permit compartmentalization of hybrid
+code in user mode.
```
rwatson, line 1518: I wonder if it’s worth spelling this out a bit more -- e.g., explaining why this would work for hybrid code?

Otherwise I think this all looks good!

bsdjhb: I've added a sentence to say what the relocation would be, not quite sure if that's what you are looking for?  I'm somewhat hesitant to speculate what implementation form relocation might take in the future though could describe that if desired?

arichardson (approved):

PR #55, opened 2023-07-06
Deprecate or remove several CHERI-RISC-V instructions

bsdjhb:
I removed CSub since the toolchain has already effectively removed it.  I deprecated the other instructions.  I did not remove references to CGetAddr in the generic architecture chapter.


insn-riscv/cgetaddr.tex
```diff
@@ -14,3 +14,7 @@ \subsubsection*{Format}
 \end{center}
 
 \sailRISCVisarefbody{CGetAddr}
+
+\subsubsection*{Deprecated}
```
arichardson, line 18: CGetAddr has also been removed from the compiler, it now emits addi 0 instead.

arichardson: Yes, it will expand it to an `ADDI 0` (`MV`).

bsdjhb: Ok, I've updated the series to remove CGetAddr entirely.

arichardson (approved):

PR #60, opened 2023-07-11
riscv: Update the main CHERI-RISC-V chapter for mode-dependent jumps.

bsdjhb:
This does not explicitly mention JALR.DDC and JALR.CAP.

### DIFF
--- a/chap-cheri-riscv.tex
+++ b/chap-cheri-riscv.tex
@@ -231,7 +231,7 @@ The following changes are specific to CHERI-RISC-V:
 \item The \cflags{} field contains a single bit indicating the ``capability
   encoding mode'' to use when the capability is installed as \PCC{}.
 \item In the non-compressed RISC-V encoding, the capability encoding mode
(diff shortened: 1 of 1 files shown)


bsdjhb:
I am going to work on a followup to this that overhauls some of the language about capmode in general, but this is probably worth fixing first.  The compressed section probably needs more updates after this still, but that might be better to wait until my capmode cleanup.

chap-cheri-riscv.tex
```diff
@@ -684,17 +676,18 @@ \subsubsection{Encoding Modes}
 
 \begin{description}
 \item[Integer encoding mode (0)] Conventional RISC-V execution mode, in which
-  address operands to existing RISC-V load and store opcodes contain
+  address operands to existing RISC-V load, store, jump, and \insnnoref{auipc} opcodes contain
```
jrtc27, line 679:
```suggestion
  address operands to existing RISC-V load, store, jump, and \insnnoref{AUIPC} opcodes contain
```

arichardson (approved):
Looks sensible

PR #61, opened 2023-07-13
Add an additional paragraph to the 64-bit CHERI format description.

bsdjhb:
Also add a note in the 9.0 change list.

### DIFF
--- a/chap-architecture.tex
+++ b/chap-architecture.tex
@@ -1186,7 +1186,7 @@ It is a development from the CHERI-128 compression format described in \cref{app
 For a more detailed rational behind some of the encoding decisions see \cref{sec:rational:comressed}.
 
 \Cref{fig:cheric128} shows the capability format and decoding method for 128-bit CHERI concentrate.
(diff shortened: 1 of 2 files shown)


chap-architecture.tex
```diff
@@ -1474,6 +1475,11 @@ \subsubsection{CHERI Concentrate 64-bit format for 32-bit address spaces}
 \vspace{-1.5em}
 \end{figure}
 
+\Cref{fig:cheric64} shows the capability format and decoding method
+for 64-bit CHERI concentrate.  The format contains a 32-bit address,
+$a$, 12 hardware defined permission bits, a 4-bit object type and 15
```
bsdjhb, line 1480: I don't know if the 15 bits is correct.  From the diagram it looks like B is 8 bits, T is 6 bits, and IE is 1 bit == 15, but with 16 bits for p + otype, that leaves an extra bit unaccounted for.  In the source it seems that IE is rendered as a bitbox of 2 bits, so is that really 2 bits or is there a 1-bit reserved field missing (that is used on RISC-V for the flags field)?

PR #74, opened 2023-07-18
Update xCCSR for V9

bsdjhb:


chap-cheri-riscv.tex
```diff
@@ -472,21 +472,26 @@ \subsubsection{Capability Control and Status Registers (CCSRs)}
 \begin{figure}[!h]
 \begin{center}
 \begin{bytefield}[bitwidth=\textwidth/34]{32}
-  \bitheader[endianness=big]{0,1,2,31} \\
-  \bitbox{32}{\textbf{WIRI}}
+  \bitheader[endianness=big]{0,30,31} \\
+  \bitbox{1}{\texttt{tc}}
+  \bitbox{1}{\rotatebox{90}{\texttt ndr}}
```
bsdjhb, line 477: "ndr" doesn't easily fit in a single-bit box, so I rotated the label 90 degrees.

arichardson: That sounds reasonable. The only other thing I could think of is ~D~, but that looks weird.

Should we reserve the "No PCC relocation bit" here as well?

bsdjhb: Oh, I didn't know you envisioned a separate bit for PCC.  If you intend a separate bit, then we should add it, yes.  Alternatively, we could use a single 'NR' ("no relocation") bit (which would also fit in the box) to mean that both DDC and PCC relocation is disabled.  I don't think we expect the feature bit here to really be independent for DDC and PCC?

arichardson: I'd be happy to have a single bit - I just used two bits for the two separate PRs.

bsdjhb: Ok, I've renamed the bit to `nr` in this version.

PeterRugg (approved):
Looks good, and I've started implementing these in HW so I can align with sail. It looks like sail still has the enable bit at the bottom of CCSR (although it's now scrapped dirty): is that specified anywhere or do we imagine it will go away? We could also add a comment to the effect of highlighting that the various bits track major changes we've made to the SPEC rather than all configurations being sensible long-term? May be too rationale-ish though.

chap-cheri-riscv.tex
arichardson: Done in https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/77

bsdjhb:
> Looks good, and I've started implementing these in HW so I can align with sail. It looks like sail still has the enable bit at the bottom of CCSR (although it's now scrapped dirty): is that specified anywhere or do we imagine it will go away? We could also add a comment to the effect of highlighting that the various bits track major changes we've made to the SPEC rather than all configurations being sensible long-term? May be too rationale-ish though.

The dirty bit was removed from the ISA spec when removing support for split register files in #10.  The enable bit was moved to `menvcfg` and `senvcfg` in #22.  I don't think any implementations have yet been updated for #22.  In particular, `Xenvcfg` is post 1.11 of the privilege spec IIRC (though is currently ratified in the latest privilege spec)

PR #78, opened 2023-07-21
RISC-V: Trim language around the CHERI ISA being a design space.

bsdjhb:
CHERI-RISC-V's definition is more concrete with fewer alternatives than when the specification was first written.  The only remaining design space choice is 32-bit vs 64-bit which is probably not enough to justify the remaining language.  This does still keep the idea of a design space in the list of goals motivating the initial desire to apply CHERI to RISC-V as that list is largely historical.

### DIFF
--- a/chap-cheri-riscv.tex
+++ b/chap-cheri-riscv.tex
@@ -12,10 +12,6 @@ elaborating CHERI capabilities within a specific architecture: 32-bit
 and 64-bit RISC-V.
 Wherever possible, CHERI-RISC-V implements the architecture-neutral concepts
 described in Chapter~\ref{chap:architecture}.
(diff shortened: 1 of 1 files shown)


bsdjhb:
I suspect Robert is unlikely to review this before the end of the month (when we'd like to have V9 ready).  I believe this is inline with the pitch we wish to use for V9 though.

brooksdavis (approved):
This makes sense to me. There are certainly some open questions around things like compartmentalization mechanisms, final encodings, etc but we've largely made the big decisions around register files, exceptions, etc.

PeterRugg:
FWIW, when I was reading through, this section doesn't make sense (the list of one item is referred to in the plural) so can't be left as is. I would definitely just remove it.

PR #81, opened 2023-07-25
Enable an existing section describing CHERI-RISC-V as non-standard ex…

bsdjhb:
…tension

This section was written by Prashanth a long while ago but has been commented out since it was written.

### DIFF
--- a/chap-cheri-riscv.tex
+++ b/chap-cheri-riscv.tex
@@ -242,28 +242,27 @@ the RISC-V instruction set.
 Instruction opcode encodings can be found in
 Appendix~\ref{app:isaquick-riscv}.
 
-\pmnote{Perhaps mention something like this:
-
 \subsection{CHERI as a non-standard RISC-V extension}
 
 CHERI is integrated into the RISC-V ISA as a non-standard extension
 named Xcheri, and follows the idioms for RISC-V extensions to the
 extent possible.  In the extension terminology of the RISC-V
-specification, CHERI is a \emph{greenfield} extension since it adds
-new instructions by populating a new instruction encoding space.  The
+specification, CHERI is mostly a \emph{greenfield} extension since it adds
+most new instructions by populating a new instruction encoding space.  The
 prefix used for the encoding is currently ``1011011'', placing it in
 the \emph{custom-2/rv128} opcode space that the specification allows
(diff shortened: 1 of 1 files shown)


bsdjhb:
It would be nice to either fix any issues in this section and include it, or trim it.

chap-cheri-riscv.tex
```diff
@@ -260,7 +258,6 @@ \subsection{CHERI as a non-standard RISC-V extension}
 hardwired to 1 on boot to indicate the presence of a non-standard
 extension.  Information tying this set X bit to the Xcheri extension
 would be communicated to system software in a platform-specific manner.
-}
 \pdrnote{Agreed: this would be great, although pedantically, our adding of
```
bsdjhb, line 261: In particular, Peter's note here probably needs resolving one way or another?

bsdjhb: Humm, it seems like while most new instructions are in a new prefix, LC/SC and the atomics are both "brownfield" as you say.  I will try to describe this more fully.

chap-cheri-riscv.tex
```diff
@@ -241,26 +241,27 @@ \section{CHERI-RISC-V Specification}
...
 the \emph{custom-2/rv128} opcode space that the specification allows
 for use for custom instruction set extensions on RV64; this makes it a
-standard-compatible global encoding. (See however the discussion in
-Section~\ref{section:cheri-risc-v-rv128-lq-sq}.)
+standard-compatible global encoding.  However, we also propose a few
+new instructions in existing encoding ranges.  The new instructions to
+load and store capabilities are \emph{brownfield} extensions to the
+LOAD and STORE opcodes in the base integer ISAs.  In addition, CHERI
+adds new atomic operation instructions which are \emph{brownfield}
+extensions to the AMO opcode.
```
bsdjhb, line 259: I'm not quite sure it strictly matches the strict language in the base RISC-V spec to say that CHERI is both a greenfield extension and two brownfield extensions, but I think it is a more accurate description

PeterRugg (approved):

chap-cheri-riscv.tex
PeterRugg: Great! Happy to delete my comment when I read through, or feel free to delete it now.

PR #88, opened 2023-08-04
Update future architecture plans section

PeterRugg:
Some of the suggestions in the section no longer make sense or have been completed. I've tried to bring it up to date as conservatively as possible.

### DIFF
--- a/chap-rationale.tex
+++ b/chap-rationale.tex
@@ -261,6 +261,18 @@ Further, we felt that the working-set properties of the two register files
 might be quite different; effectively pinning the two to one another would
 reduce the efficiency of both.
 
+With a split register file, it may be useful for the architecture to add
+instructions for copying non-capability data from a capability register
+into a general-purpose integer register. A use case is when a function is called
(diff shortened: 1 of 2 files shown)


chap-rationale.tex
```diff
@@ -261,6 +261,18 @@ \section{Capability Register File}
 might be quite different; effectively pinning the two to one another would
 reduce the efficiency of both.
 
+With a split register file, it may be useful for the architecture to add
```
bsdjhb, line 264: This change is fine, but I feel this section is now a bit out-dated overall as it still claims the use of split register files in general.  I hope I didn't add a separate rationale section on split vs merged in my earlier changes.  If so, that section and this need to be merged into one probably.

bsdjhb (approved):

chap-rationale.tex
PeterRugg: Yes, that's true. This section definitely hints at us still using split register files, but a tiny bit of wordsmithing could probably change that, as it mostly all still applies.

PR #89, opened 2023-08-04
Update DARPA acks for ISAv9

brooksdavis:


bsdjhb (approved):

acknowledgments.tex
```diff
@@ -3,7 +3,7 @@ \section*{Acknowledgments}
 \vspace{-0.2cm}
 
 The authors of this report thank members of the CTSRD, MRC2, ECATS,
-CIFV, and REMS teams, our past and current research collaborators at SRI and
+CIFV, REMS, ETC, and MTSS teams, our past and current research collaborators at SRI and
 Cambridge, as well as colleagues at other institutions who have provided
 invaluable feedback and continuing support throughout this work:
```
bsdjhb, line 9: Do we also need to add more people to the list of people below?  I think at least @gvnn3 is missing?

brooksdavis: Hmm, not sure how to generate that list. I guess reading meeting attendees might work?

bsdjhb (approved):

PR #90, opened 2023-08-04
RISC-V Sail update

bsdjhb:


chap-isaref-riscv.tex
```diff
@@ -117,34 +122,40 @@ \subsection*{Types used in function definitions}
...
 The overloaded function \isail{F(n)} is used to read or write floating-point register \isail{n}.
 
 \medskip
 \sailRISCVval{memBitsToCapability}
 \sailRISCVval{capToMemBits}
 \sailRISCVval{int\_to\_cap}
+\sailRISCVval{mem\_read\_cap}
 
 \medskip
+\sailRISCVval{ddc\_and\_resulting\_addr}
```
bsdjhb, line 158: This function could possibly use a description (maybe ideally in Sail that auto-populates here?)

arichardson: There is a doc comment, so sail should generated the docs for this?

arichardson: Ah I think this is the LaTex output issue where the doc comment needs to be on a `val` and the `func` comment is ignored.

PR #94, opened 2023-08-07
chap-rationale: Rewrite the capability register file section

bsdjhb:
Shift the narrative of this section a bit more firmly to describe the separate register file on MIPS as the starting point and keep most of the existing text but framed as a series of steps that led to the merged register file approach currently used.  While here, also update some text in this section that speculated about adding CGetAddr to note that that CGetAddr was in fact added and explicitly reference it by name.

### DIFF
--- a/chap-rationale.tex
+++ b/chap-rationale.tex
@@ -206,15 +206,12 @@ properties.
 
 \section{Capability Register File}
 
-The decision to separate the capability-register file from the general-purpose
-integer
-register file is somewhat arbitrary from a software-facing perspective: we
-envision capabilities gradually displacing general-purpose integer registers as
-pointers, but where management of the two register files will remain largely
-the same, with stack spilling behaving the same way, and so on, as is already
-the case for disjoint integer and floating-point register files.
+CHERI extends existing general-purpose integer registers to hold
+capabilities.  This design is similar to the manner in which the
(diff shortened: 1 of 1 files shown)


chap-rationale.tex
```diff
@@ -229,25 +226,23 @@ \section{Capability Register File}
...
+instructions for copying non-capability data from a capability register
+into a general-purpose integer register. A use case for this was when a function was called
+with a parameter whose type is the union of a pointer and a non-pointer type,
+such as an int. This parameter had to be passed in a capability register, because
+the tag needed to be preserved when it held a capability. If the body of
+the function accessed the non-capability branch of the union, it needed to
+get the non-capability bits out of the capability register and into a general
+purpose register. This was originally done by spilling the capability register to the
+stack and then reading it back into a general-purpose integer register, but the register
+to register copy of \insnnoref{CGetAddr} proved faster.
```
brooksdavis, line 245: Do we want to mention the CSetAddr on the NULL register here as well?

brooksdavis (approved):
A nice improvement in decisiveness

chap-rationale.tex
bsdjhb: Perhaps.  This is definitely the place to note the use cases that motivated it.  I'm not sure I know that particular bit of history well enough to describe it.

PR #106, opened 2023-08-14
Use alphabetic bibliography style

bsdjhb:


brooksdavis:
Looks good to me. It won't be 100% stable, but you'll only see changes in small areas like if we added another 2020 Arm citation that sorted in the middle.
image

brooksdavis:
That screen shot did not work as expected...

brooksdavis (approved):
We might separately want to bump URL dates at some point.

PR #122, opened 2024-01-23
Added text extension about isentries

francislaus:
This is the result of research on isentries in the past weeks and months.

### DIFF
--- a/app-experimental.tex
+++ b/app-experimental.tex
@@ -1355,7 +1355,8 @@ Isentries are not designed for secrecy about where a compartment wants to jump, 
 Similar methods, e.g., cinvoke gives out sealed code and data capabilities, which are then invoked as a pair.
 The code capability itself is not secret.
 
-It is important to note that the linked pair of a \insnnoref{CLIL} and \insnnoref{CJAURL} instruction does not give any guarantees about memory consistency. Memory consistency is not an issue tackled by CHERI, but needs to be addressed by other architectural mechanisms, e.g. atomic memory operations.
+It is important to note that in the current state of this proposal the linked pair of a \insnnoref{CLIL} and \insnnoref{CJAURL} instruction does not give any guarantees about memory consistency.
+Further work could determine how the reservation interacts with the memory model.
 
 \subsubsection{Allowing more instructions in between \insnnoref{CLIL} and \insnnoref{CJAURL}}
 
--- a/app-experimental.tex
+++ b/app-experimental.tex
@@ -1468,12 +1468,42 @@ sealed.  Because these capabilities reside in memory, the instruction
 constructing these ``points to pair'' indirect sentries likely cannot perform
 any validation on their contents.
 
+We allocate the following \cotype{} $2^{64} - 6$ from the reserved otype space for ``points to pair'' isentries.
+%Furthermore, we allocate the \cotype{} $2^{64} - 7$ for capabilities that have been loaded through ``points to pair'' linked isentry loads.
+
 As described in the previous section about ``points to PCC'' isentries, an
 instruction facilitating all operations is complex and imposes the need for
 micro-operations. Thus, we also propose to split the ``points to pair'' isentry
-instruction into a pair of linked instructions.
+instruction into multiple linked instructions.
+There are four operations included in jumping to a ``points to pair'' isentry: loading the data capability, loading the code capability, jumping to the code capability, and installing the data capability in the IDC register.
+All these operations include capability checks.
+We propose the following three instructions for the four operations above:
+
+\begin{enumerate}
+  \item CLILC: CHERI load isentry linked code: this instruction has two operands. The register index of the ``points to pair'' isentry and a destination register index.
+  The instruction loads the capability at the address the isentry is pointing and stores it in the destination register.
+  The loaded capability is sealed with the 
+  Furthermore it creates a reservation about the linked load or it manipulates an already existing reservation.
+  The instruction does the following capability checks:
+  \begin{itemize}
+    \item Check whether the isentry is tagged.
+    \item Check whether the isentry is sealed with the ``points to pair'' isentry otype.
+    \item Check whether the isentry can load a CAP-WIDTH wide word.
+  \end{itemize}
+  \item CLILD: CHERI load isentry linked code; this instruction has one operand, which is the register of the ``points to pair'' isentry.
+  The implicit destination register is IDC.
+  The instruction loads the capability at the address of (isentry + CAP WIDTH) and stores it in the implicit destination register.
+  Furthermore it creates a reservation about the linked load or it manipulates an already existing reservation.
+  The instruction does the following capability checks:
+  \begin{itemize}
+    \item Check whether the isentry is tagged.
+    \item Check whether the isentry is sealed with the ``points to pair'' isentry otype.
+    \item Check whether the isentry can load a CAP-WIDTH wide word.
+  \end{itemize}
+\end{enumerate}
 
 ``Points to pair'' isentries need a separate instruction to seal the capability.
+This instruction will use the \cotype{} $2^{64} - 6$.
 
 % >>>
 % >>>
--- a/app-experimental.tex
+++ b/app-experimental.tex
@@ -1329,21 +1329,21 @@ being written with valid capabilities or none of them being written.
 
 \subsubsection{Definition of linked pair of instructions}
 
-In this extension, we propose to split the cisentryj (CHERI indirect sealed entry jump) instruction into two sub instructions: Lil (load isentry linked) and cjaurl (CHERI jump and unseal into register linked).
+In this extension, we propose to split the \insnnoref{CISentryJ }(CHERI indirect sealed entry jump) instruction into two sub instructions: \insnnoref{CLIL} (CHERI load isentry linked) and \insnnoref{CJAURL} (CHERI jump and unseal into register linked).
 In the following paragraphs, we will explain what these instructions do and how they enforce the security model we are designing.
 
-The lil instruction loads a capability through an isentry capability and clears its tag.
-In the subsequent instruction this capability can be picked up by a cjaurl, which then jumps to this code capability and unseals the isentry into ct6.
+The \insnnoref{CLIL} instruction loads a capability through an isentry capability and clears its tag.
+In the subsequent instruction this capability can be picked up by a \insnnoref{CJAURL}, which then jumps to this code capability and unseals the isentry into ct6.
 The two instructions are linked and all security properties are held as explained in the following paragraphs.
 
-If attackers executed only the lil instruction, this will leave them with an untagged code capability.
+If attackers executed only the \insnnoref{CLIL} instruction, this will leave them with an untagged code capability.
 This code capability cannot be used in regular jumps because it is untagged.
-The only way it can be used for a jump is in a subsequent cjaurl instruction.
+The only way it can be used for a jump is in a subsequent \insnnoref{CJAURL} instruction.
 
 An attacker can only jump to a code capability with the isentry it came from.
-This can be divided into two cases: If the instruction following the lil is a cjaurl, this will jump to the code pointed to by the code capability and unseal the isentry.
-The register holding the isentry is stored in the reservation information of the lil instruction.
-In the second case, if the instruction following the lil is not a cjaurl, the reservation will have become invalid.
+This can be divided into two cases: If the instruction following the \insnnoref{CLIL} is a \insnnoref{CJAURL}, this will jump to the code pointed to by the code capability and unseal the isentry.
+The register holding the isentry is stored in the reservation information of the \insnnoref{CLIL} instruction.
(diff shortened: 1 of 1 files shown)


app-experimental.tex
```diff
@@ -1311,6 +1315,69 @@ \subsection{Points-to-PCC} % <<<
...
+code capability and unsealing the data capability.
+This property of having linked will retain the atomicity of the both IDC and PCC
+being written with valid capabilities or none of them being written.
+
+\subsubsection{Definition of linked pair of instructions}
+
+In this extension, we propose to split the cisentryj (CHERI indirect sealed entry jump) instruction into two sub instructions: Lil (load isentry linked) and cjaurl (CHERI jump and unseal into register linked).
+In the following paragraphs, we will explain what these instructions do and how they enforce the security model we are designing.
+
+The lil instruction loads a capability through an isentry capability and clears its tag.
```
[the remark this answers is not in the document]
PeterRugg: Boom, we were discussing, and I reckon we can make this a lot nicer. Instead of having this loaded code cap being "sort of untagged but sort of not", we can just have a dedicated otype that says "only usable if it's being used in a cjaurl as the code cap with a valid reservation". When someone sets up one of these objects, they seal the code cap with that otype, and then because it's sealed it can't be used by anyone for anything else. That will also prevent the revoker from missing it, and presumably all sorts of other problems that we hadn't thought of yet due to these partially untagged values

PeterRugg (changes requested):
Looks good! A few changes to discuss tomorrow perhaps.

app-experimental.tex
```diff
@@ -1311,6 +1315,69 @@ \subsection{Points-to-PCC} % <<<
...
+A write to ct6 -- where the code capability is stored -- will equally lead to the reservation being invalidated.
+
+There always is only ever one reservation.
+A lil instruction is a write to the ct6 register and thus invalidates the reservation.
+A read of any register does not invalidate the reservation.
+None of the information is secret and the reservation will still point to a correct copy of the isentry as well as the code capability.
+
+In this proposal there is no maximum number of instructions between the lil and cjaurl instruction.
+
+
```
nwf, line 1380: What happens if the core takes an interrupt between `lil` and `cjaurl`?  Is it possible to spill and restore these reservations?  (That is, how does context switching work?)

jrtc27: Or you take a page fault (whether because you put a memory instruction in between lil and cjaurl, or because those two instructions straddled a code page boundary)?

jrtc27: (and I wouldn't be thrilled about the OS being able to forge such reservations... it effectively becomes a somewhat-janky capability forgery oracle)

PeterRugg: Ah, when we discussed this, I think the idea was that it can fail. You have to put it in a loop like LR SC.

jrtc27: Then you can't have an arbitrary, or even just unbounded, set of instructions in between, because you need a forward progress guarantee

PeterRugg: Yep, we discussed the pattern being roughly: LIL, busy work (presumably register clearing), CJAURL, LIL, CJAURL, unconditional jump back 2. The first two are optimistic, but the last 2 are the only ones that would need a progress guarantee.

app-experimental.tex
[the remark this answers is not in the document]
PeterRugg: Hmm, it helps a little, but the nefarious code could still swap out the data capability, so you still need something like a reservation, it seems.

app-experimental.tex
francislaus: I should have phrased it differently. I was aware of the need for forward progress guarantee, but I am not yet sure what a sensible number of instructions is. I will have a think about what a good number is. Having a forward progress guarantee only if the cjaurl follows a lil instruction could incentivise microarchitects to optimise for that which would harm the case of pushing the lil before doing work and thus attempting to mitigate a potential load to use penalty.

app-experimental.tex
```diff
@@ -1311,6 +1315,69 @@ \subsection{Points-to-PCC} % <<<
...
+Because of this linkage, the attacker can never use a code capability and a data capability together, which have not been intended to work together.
+Thus the linking keeps the atomicity of the instruction.
+
+Loading the code capability does not break CHERI confidentiality.
+Isentries are not designed for secrecy about where a compartment wants to jump, but to implement jumping to a sealed pair of capabilities.
+Similar methods, e.g., cinvoke gives out sealed code and data capabilities, which are then invoked as a pair.
+The code capability itself is not secret.
+
+It is important to note that the linked pair of a lil and cjaurl instruction does not give any guarantees about memory consistency. Memory consistency is not an issue tackled by CHERI, but needs to be addressed by other architectural mechanisms, e.g. atomic memory operations.
```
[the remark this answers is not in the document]
francislaus: The reason for this paragraph to exist is to avoid confusing people who think this `ll` - `sc` like concept. I agree the consequences could be interesting. Let's have a chat about that later. For now, I will include your changes:)

app-experimental.tex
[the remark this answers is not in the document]
PeterRugg: I see you've added this. Looks good. I wonder about sealing on the load path though. Should be fine, but seems slightly messier than just saying whoever installs the code cap had to seal it. I guess that requires yet another new instruction though, or messing with csealentry to allow sealing with the new type as well

PeterRugg (approved):

app-experimental.tex
```diff
@@ -1387,14 +1460,96 @@ \subsection{Points to Pair} % <<<
...
+micro-operations. Thus, we also propose to split the ``points to pair'' isentry
+instruction into multiple linked instructions.
+There are four operations included in jumping to a ``points to pair'' isentry: loading the data capability, loading the code capability, jumping to the code capability, and installing the data capability in the IDC register.
+All these operations include capability checks.
+We propose the following three instructions for the four operations above:
+
+\begin{enumerate}
+  \item \insnnoref{CLILC}: CHERI load isentry linked code: this instruction has two operands. The register index of the ``points to pair'' isentry and a destination register index.
+  The instruction loads the capability at the address the isentry is pointing and stores it in the destination register.
+  The loaded capability is sealed with the
```
PeterRugg, line 1485: sentence needs finishing?

app-experimental.tex
```diff
@@ -1387,14 +1460,96 @@ \subsection{Points to Pair} % <<<
...
+  The implicit destination register is IDC.
+  The instruction loads the capability at the address of (isentry + CAP WIDTH) and stores it in the implicit destination register.
+  Furthermore it creates a reservation about the linked load or it manipulates an already existing reservation.
+  The instruction does the following capability checks:
+  \begin{itemize}
+    \item Check whether the isentry is tagged.
+    \item Check whether the isentry is sealed with the ``points to pair'' isentry otype.
+    \item Check whether the isentry can load a CAP-WIDTH wide word.
+  \end{itemize}
+  \item \insnnoref{CJAURLP}: CHERI jump and unseal into register linked pair; this instruction has one three operands of which one is explicit and the other two are implicit.
```
PeterRugg, line 1503: "has one three operands"

app-experimental.tex
```diff
@@ -1387,14 +1460,96 @@ \subsection{Points to Pair} % <<<
...
+
+%\subsubsection{Order of instructions}
+%
+%This section describes one way of implementing the three instructions and its implications.
+
+\subsection{Encoding of isentry instructions}
+
+We specified multiple instructions in this Chapter.
+Some of them exhibit similar functionality.
+Therefore, we propose to encode the following instructions in the same encoding space and use ``mode'' bits to distinguish between the respective behaviour.
```
PeterRugg, line 1545: The mode bits are encoded in the instruction rather than dynamic in some CSR, I take it?

PeterRugg:
Looks like a sensible generalisation to points-to-pair to me!

app-experimental.tex
francislaus: @PeterRugg I was thinking to encode them similar to branch instructions in RISC-V. For example, all sealing instructions would have the same general behaviour, but the exact sealing value would be dependent on the mode bits. `00` could be sentries, `01` could be points to PCC isentries, `02` could be points to pair isentries.

francislaus: Would that be a sensible approach in your opinion? Or would rather go through unused `funct5` values?

francislaus:
@PeterRugg Thank you so much!:)

app-experimental.tex
```diff
@@ -1311,6 +1315,73 @@ \subsection{Points-to-PCC} % <<<
...
+micro-operations in microarchitectures. Thus we propose to have the two operations
+of the complex \insnnoref{CInvokeInd} to be split into a linked pair of instructions:
+an instruction for load of the code capability and an instruction for jumping to the
+code capability and unsealing the data capability.
+This property of having linked will retain the atomicity of the both IDC and PCC
+being written with valid capabilities or none of them being written.
+
+\subsubsection{Definition of linked pair of instructions}
+
+In this extension, we propose to split the \insnnoref{CISentryJ }(CHERI indirect sealed entry jump) instruction into two sub instructions: \insnnoref{CLIL} (CHERI load isentry linked) and \insnnoref{CJAURL} (CHERI jump and unseal into register linked).
```
kwitaszczyk, line 1332: Is `CISentryJ` anywhere defined? I couldn't find it.

app-experimental.tex
```diff
@@ -1311,6 +1315,73 @@ \subsection{Points-to-PCC} % <<<
...
+In the following paragraphs, we will explain what these instructions do and how they enforce the security model we are designing.
+
+The \insnnoref{CLIL} instruction loads a capability through an isentry capability and seals it.
+For this, we use the \cotype{} $2^{64} - 5$, which is reserved.
+In the subsequent instruction this capability can be picked up by a \insnnoref{CJAURL}, which then jumps to this code capability and unseals it as well as unsealing the isentry into ct6.
+The two instructions are linked and all security properties are held as explained in the following paragraphs.
+
+If attackers executed only the \insnnoref{CLIL} instruction, this will leave them with a sealed code capability.
+This code capability cannot be used in regular jumps because it is seald with the \cotype{} $2^{64} - 5$.
```
kwitaszczyk, line 1341: Missing "e" in "sealed".

app-experimental.tex
```diff
@@ -1311,6 +1315,73 @@ \subsection{Points-to-PCC} % <<<
...
+In the second case, if the instruction following the \insnnoref{CLIL} is not a \insnnoref{CJAURL}, the reservation will have become invalid.
+Thus the attacker cannot jump to the code capability.
+Equally, the isentry has never been unsealed and thus cannot be used by the attacker.
+
+Because of this linkage, the attacker can never use a code capability and a data capability together, which have not been intended to work together.
+Thus the linking keeps the atomicity of the instruction.
+
+Loading the code capability does not break CHERI confidentiality.
+Isentries are not designed for secrecy about where a compartment wants to jump, but to implement jumping to a sealed pair of capabilities.
+Similar methods, e.g., cinvoke gives out sealed code and data capabilities, which are then invoked as a pair.
```
kwitaszczyk, line 1356:
```
\insnnoref{CInvoke}
```

app-experimental.tex
```diff
@@ -1387,14 +1460,97 @@ \subsection{Points to Pair} % <<<
...
+%This section describes one way of implementing the three instructions and its implications.
+
+\subsection{Encoding of isentry instructions}
+
+We specified multiple instructions in this Chapter.
+Some of them exhibit similar functionality.
+Therefore, we propose to encode the following instructions in the same encoding space and use ``mode'' bits to distinguish between the respective behaviour.
+
+\begin{itemize}
+  \item \insnref{CSealEntry}, \insnnoref{CISealPCC}, and \insnnoref{CISealPair}
```
kwitaszczyk, line 1549: Are you proposing to define `CISealPCC` and `CISealPair` as well? Do they intend to construct isentries similarly to `CSealEntry`? Would they modify the mode bits as described above?

app-experimental.tex
```diff
@@ -1387,14 +1460,97 @@ \subsection{Points to Pair} % <<<
...
+PCC and IDC.  Invocation of such an isentry performs two capability loads through
 an ephemeral (architecturally invisible), unsealed copy of the given sentry and
 then, with both capabilities in hand, installs both into the register file
 atomically.  There is no requirement that the two capabilities pointed at be
 sealed.  Because these capabilities reside in memory, the instruction
 constructing these ``points to pair'' indirect sentries likely cannot perform
 any validation on their contents.
 
+We allocate the following \cotype{} $2^{64} - 6$ from the reserved otype space for ``points to pair'' isentries.
+%Furthermore, we allocate the \cotype{} $2^{64} - 7$ for capabilities that have been loaded through ``points to pair'' linked isentry loads.
```
kwitaszczyk, line 1472: Shouldn't it be uncommented? `$2^{64} - 7$` is used below.

PR #123, opened 2024-03-18
Thread identification

francislaus:
This is the first summary of how thread identification can help with compartmentalistation

### DIFF
--- a/app-experimental.tex
+++ b/app-experimental.tex
@@ -2693,6 +2693,114 @@ as well as its internal operation, are left to future work.
 
 % >>>
 % >>>
+\section{Thread Identification} % <<<
+
+\subsection{Motivation}
+
+Compartmentalisation models often rely on trusted code performing the switch from one compartment to another.
+This trusted code needs a mechanism to retrieve a data capability, e.g., for a stack.
+When using sentries (sealed entry capabilities) to call into trusted code, the same piece of trusted code can be called from multiple threads and each thread must have a distinct trusted data capability.
+Thus, we have the need for reliable thread identification.
+This can be solved by calling into the entity that provides threading support.
+In the conventional case, this would mean a call to kernel space at every time trusted code is entered.
+Thus, we need to have a way of both fast and reliably determining the current thread ID (TID).
+
+The TID is not the same as Thread-Local Storage (TLS).
+When jumping from one compartment to another, the TLS should change per compartment, but the TID should remain the same over all calls.
+The commonly used RISC-V calling convention defines \texttt{x4} as the thread pointer register.
+We cannot use \texttt{x4} as the TID register because it is not a reliable source.
+Any compartment can manipulate it.
+
+\subsection{Design}
+
+We currently envision multiple layers to this design.
+
+\begin{enumerate}
+\item The most simple design is to have one thread in the system called the Supervisor Thread ID (STID).
+      The register is a CSR that is only writable from S mode or more privileged modes if the ASR permission bit is set.
+      It is exposed as a read-only register to U mode, which is called User Thread ID (UTID).
+      We propose the following allocation in the CSR space: STID: \texttt{0x541} (RW), UTID: \texttt{0xC30} (RO)
+\item The design above can be extended by tagging the integer state in the TID register.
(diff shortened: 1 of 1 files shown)


PeterRugg (approved):

app-experimental.tex
```diff
@@ -2693,6 +2693,125 @@ \section{Distributed Capabilities For Peripherals And Accelerators} % <<<
 
 % >>>
 % >>>
+\section{Thread Identification} % <<<
+
+\subsection{Motivation}
+
+Compartmentalisation models often rely on trusted code performing the switch from one compartment to another.
+This trusted code needs a mechanism to retrieve a data capability, e.g., for a stack.
+When using sentries (sealed entry capabilities) to call into trusted code, the same piece of trusted code can be called from multiple threads and each thread must have a distinct trusted data capability.
```
rwatson, line 2702: I think more properly, it’s that each thread must have a distinct stack -- even if it’s reached via the same trusted data capability. But they will need to select the right stack.

app-experimental.tex
```diff
@@ -2693,6 +2693,125 @@ \section{Distributed Capabilities For Peripherals And Accelerators} % <<<
...
+The commonly used RISC-V calling convention defines \texttt{x4} as the thread pointer register.
+We cannot use \texttt{x4} as the TID register because it is not a reliable source.
+Any compartment can manipulate it.
+
+\subsection{Design}
+
+We currently envision multiple layers to this design.
+
+\begin{enumerate}
+\item The most simple design is to have one thread in the system called the Supervisor Thread ID (STID).
```
rwatson, line 2719: “In the system” seems a very broad scope. Do we mean per core / hardware context / HART / something like that?

app-experimental.tex
rwatson: Hmm. Maybe in fact my comment is more broad in scope: “trusted data capability” is not a term we’ve defined previously, and may require some discussion. Note that in, for example, an otype-type model, you may already have a trustworthy data capability -- you just need a thread ID to index it so you can load the right per-thread stack capability. Whereas in other models, such as executive-restrictive, you need a dedicated stack pointer capability and don’t need to derive that from another capability.

app-experimental.tex
```diff
@@ -2693,6 +2693,125 @@ \section{Distributed Capabilities For Peripherals And Accelerators} % <<<
...
+\subsection{Implementation Notes}
+
+Access to CSR registers is usually slow in microarchitectures because many CSRs take effect on the system and thus need to stall the pipeline or even need to redirect.
+The TID and TDC registers are different because they are pure-data registers and no value stored in it has any effect on other architectural state.
+Thus, we propose to treat these registers differently.
+
+We note that the TID register is written when every there is a thread switch, but not otherwise.
+However, the TID register is read on every trampoline entry.
+Thus, we conclude that the reading speed of the TID register matters for overall performance whereas the writing speed does not.
+We envision that this strategy might be useful to other pure-data CSRs.
```
rwatson, line 2813: Perhaps it would be useful to better characterise / discuss the performance considerations here. In the CheriBSD library compartmentalisation model, you’d need to context switch TID every time you context switch a user thread, but not every time you perform a domain switch within a user thread. Thread context switching speeds do matter, so we need to be cautious in the above statement: it is less performance critical, but it’s not irrelevant.

app-experimental.tex
francislaus: The term "trusted data capability" should be defined. I wanted to keep the terminology broad. Some models might not only need a trusted stack. As I understand the current form of the library-compartmentalisation model has a table of stacks for the compartments called into. This table is stored in a separate capability. One could design that to also be accessible in a thread ID indexed way.

app-experimental.tex
francislaus: I should change that to hardware thread. That also is what the RISC-V terminology seems to be.

jrtc27: Userspace (generally) isn’t dealing with hardware threads though, it’s dealing with software threads. You don’t want your software thread ID changing when you get migrated to another hart, nor do you want a different software thread to start using yours if it later runs on the same hart.

PR #124, opened 2024-05-07
CHERI ISAv10 is now a work-in-progress

bsdjhb:


cheri-architecture.tex
```diff
@@ -53,7 +53,6 @@
   \\
   SRI International, University of Cambridge, and Arm Limited
 }
-\date{22nd September 2023}
```
brooksdavis, line 56: Maybe comment out in hopes of reminding us to re-add?

brooksdavis (approved):

cheri-architecture.tex
bsdjhb: Done

PR #126, opened 2024-05-10
Added TIDC registers

francislaus:


PeterRugg (approved):
Oops, just spotted these are still up. Still worth merging, even though the SCR use (as opposed to CSR) is out of date?

francislaus:
I will merge it because I think we still want to have it in the v9/v10 architecture.

PR #130, opened 2025-12-13
Add experimental section describing design points for using DMA with CHERI

tmarkettos:
Added a section about CHERI and DMA, describing some options and setting the scene for existing two sections on physical addresses and capability colouring.

### DIFF
--- a/app-exp-dma.tex
+++ b/app-exp-dma.tex
@@ -1,0 +1,66 @@
+CHERI can be composed with DMA controllers in a number of ways.
+
+\subsection{DMA categorization}
+We can survey existing DMA controllers and discover a number of design patterns, which are helpful in understanding how CHERI can be applied to DMA subsystems.
+We outline them briefly here and refer to Markettos et al.~\cite{FIXME} for further background.
+
+\subsubsection{Transaction types}
+A DMA controller's job is to generate memory transactions based on data sources and sinks.
+These come in various forms.
+
+\emph{Memory transactions} involve the DMA controller generating a memory cycle such as a read or a write.
+
+\emph{Streaming transactions} have unidirectional data with an ordering but do not of themselves have a memory address, for example network packets received on the wire.
+These are reflected on-chip via interfaces such as AXI streaming~\cite{FIXME}, and historically off-chip via ISA-bus DMA~\cite{FIXME}  (which is still relevant for legacy PC peripherals).
+
+A DMA operation is a combination of a source and a sink, ie we can define memory-to-memory (M2M), a memory read feeds a memory write (a memory copy);
+memory-to-stream (M2S), a memory read is sent to an outgoing stream; or
+stream-to-memory (S2M), an incoming transaction on a stream generates a memory write.
+The final case (stream-to-stream, S2S) is associated with on-chip interconnect and does not concern memory access.
+
+Each case may generate a series of transactions based on data sizes and interconnect widths
+For example, a read of a 4KiB disk block from a storage controller with a 64-bit datapath might generate a stream of 512 input transactions to the DMA, which could be expanded to 1024 transactions to a 32-bit DRAM chip.
+
+Additionally between the source and sink some processing may be carried out on the data, for example RGB to YUV conversion of a video stream, making it not strictly a copy.
+
+\subsubsection{DMAs with control information in MMIO}
+DMA controllers such as the Raspberry Pi RP2350 microcontroller~\cite{FIXME, section 12.6} have all of their DMA controller state in memory-mapped I/O registers.
+There are 16 DMA channels, each one consisting of length, a Read and a Write address, which can be memory or peripheral FIFO registers.
+Once a DMA is completed an interrupt can be generated or another DMA channel can be triggered, forming basic DMA chains.
+Hence the RP2350 is M2M-only and all the control state is held in the MMIO registers, rather than memory.
+The Atmel XMEGA DMA~\cite{FIXME} operates in a similar M2M-only fashion using MMIO registers for control.
+
+%On the RP2350 specifically, four TrustZone-like security levels are provided and a memory protection unit (MPU) holds up to 8 regions to check against them.
+
+\subsubsection{DMAs with control information streamed externally}
+The AMD (Xilinx) LogiCORE IP AXI DMA~\cite{FIXME} is an example of an M2S/S2M DMA, comprising a memory access port plus streaming sources and sinks for both data and control transactions.
+In its most basic mode it can operate with transaction information in MMIO registers, as the previous category.
+However for greated performance transaction information (addresses, lengths) can alternatively be streamed into a control port.
+This enables custom logic to control and steer memory traffic.
+
+\subsubsection{Descriptor-based DMA}
+\emph{Descriptors} are small data structures typically comprising the address of a data buffer, the size of the transaction (eg number of bytes outstanding to send) and other metadata.
+For example, how to increment addresses (allowing 2D/3D-style accesses) and what to do next after this buffer is completed.
+Descriptors can be used in roughly two ways.
+First is the fetching of descriptors from memory, either main memory or a more local memory.
+Second is the streaming of descriptors on another port, for example from a separate descriptor fetching unit (as in the LogiCORE AXI DMA above, and also the Altera Modular Scatter-Gather DMA~\cite{FIXME}).
+In-memory descriptors can often be chained or used in arrays in order to form \emph{scatter-gather lists} or \emph{ring buffers} to give a higher order structure to groups of transactions.
+
+\subsection{Composing CHERI and DMA}
+DMA controllers operate outside the world of CPU-derived capabilities.
+The question arises how to compose them.
+
+\subsubsection{Level 0: DMA clears tags}
+The simplest yet safe scenario is that DMA operates outside of the world of a CHERI-enabled CPU.
+The DMA hardware is unchanged, using integer address rather than capabilities.
+Any data written by the DMA clears tags, so the DMA is unable to forge or unsafely modify valid capabilities.
+Any pre-existing peripheral device can safely operate in this manner, although this also means they will not follow the capability protection model; an IOMMU may be required to safely constrain their memory access.
+
+\subsubsection{Level 1: addresses replaced by capabiities}
+A `CHERI-aware DMA controller' may choose to replace integer addresses in descriptors or MMIO registers with capabilities.
+Memory accesses are checked against the bounds and permissions of the capability during transactions, and invalid transactions are aborted.
+Address calculation is based on the address from the capability and is otherwise unchanged.
+
+This avoides any difficulties where the DMA controller's use of addresses and the capability model's use of bounds diverge.
+For example, a DMA controller may have separate `offset' and `length' fields to deal with restarting partially-completed transactions (for example, a receive transaction where there was insufficient data received to fill the buffer);
+these would not map directly to the bounds of a capability.
--- a/app-exp-dma.tex
+++ b/app-exp-dma.tex
@@ -84,3 +84,34 @@ If the copy were to use $i$ as the inner loop instead of $j$ we may experience t
 This would be impossible to achieve while ever shrinking the bounds of the capability.
 
 That said, depending on the requirements of the DMA controller, perhaps these limitations can be accepted in order to simplify the protection model.
+The value of a single representation shared between software driver and DMA controller may outweigh the limitations.
+
+\subsubsection{Level 3: virtual addressing}
+In systems with MMUs, capabilities are virtually addressed.
+This means it is important to ensure that capabilities referring to one virtual address space do not cross over to another.
+In application software the OS and MMU restrict capabilities to their own address space, with carefully-controlled exceptions.
+For DMA, this brings several challenges.
+
+First, capabilities in software are generated in relation to a virtual address space of an application.
+This does not directly correspond to any address space visible to an I/O device.
+Second, I/O devices such as DMA controllers may be in their own I/O virtual address spaces as translated by an IOMMU.
+This means that not only are the capabilities with respect to different address spaces, but they represent two different \emph{kinds} of address spaces, potentially whose sources of truth do not overlap.
+
+There exist a number of potential solutions in this space.
+For a successful safe capability check, we need to also know within which address space a capability exists.
+This leads to roughly three design points.  Either we:
+\begin{itemize}
+\item we carry around the address space stored as bits inside the capability, or
+\item we maintain the color as external metadata that is supplied to wherever a capability needs to be checked, or
+\item we arrange such that structurally there is no ambiguity at the point of check as to which address space a capability refers.
+\end{itemize}
+
+In the first case, bits that travel within the in-memory capability are used to identify its address-space identifier (ASID), such as a 'color' (see section~\ref{FIXME}).
+In the second, the in-memory format is not changed but bits are supplied alongside, 'out of band' (e.g. as additional tag bits).
+In the third, we may use explicit or implicit information to infer the address space.
+For example, we know a particular interconnect endpoint always generates capabilities within a single address-space, allowing us to attach an ASID at the point of check.
+
+When checking an access against the capability we must both check the operation is permitted by the capability (i.e. the usual access check), and that also that the access is permitted by rules associated with the ASID.
+For example, deriving a new capability must retain the ASID of the original one (no implicit type conversion);
+any operation using two capabilities of different ASIDs is not permitted (no type mixing);
+and capabilities are not allowed to be stored in locations associated with another ASID (no leakage).
--- a/app-exp-dma.tex
+++ b/app-exp-dma.tex
@@ -115,10 +115,11 @@ This leads to roughly three design points.  Either we:
 	\item we arrange such that structurally there is no ambiguity at the point of check as to which address space a capability refers.
 \end{itemize}
 
-In the first case, bits that travel within the in-memory capability are used to identify its address-space identifier (ASID), such as a 'color' (see section~\ref{app:exp:peripherals}).
-In the second, the in-memory format is not changed but bits are supplied alongside, 'out of band' (e.g. as additional tag bits).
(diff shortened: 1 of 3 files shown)


app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+Once a DMA is completed an interrupt can be generated or another DMA channel can be triggered, forming basic DMA chains.
+Hence the RP2350 is M2M-only and all the control state is held in the MMIO registers, rather than memory.
+The Atmel XMEGA DMA~\cite{atmel-xmegadma} operates in a similar M2M-only fashion using MMIO registers for control.
+
+%On the RP2350 specifically, four TrustZone-like security levels are provided and a memory protection unit (MPU) holds up to 8 regions to check against them.
+
+\subsubsection{DMAs with control information streamed externally}
+The AMD (Xilinx) LogiCORE IP AXI DMA~\cite{amd-logicoredma} is an example of an M2S/S2M DMA, comprising a memory access port plus streaming sources and sinks for both data and control transactions.
+In its most basic mode it can operate with transaction information in MMIO registers, as the previous category.
+However for greated performance transaction information (addresses, lengths) can alternatively be streamed into a control port.
```
nwf:
```suggestion
However for greater performance transaction information (addresses, lengths) can alternatively be streamed into a control port.
```

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+This enables custom logic to control and steer memory traffic.
+
+\subsubsection{Descriptor-based DMA}
```
nwf, line 39: It may be worth splitting this discussion.  Descriptors as presented both augment the control information discussed (with strides) and can get placed in memory.  These aspects are separable.

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+Additionally, such devices may not be known at design-time any may change dynamically, thanks to hotpluggable PCIe cards and Thunderbolt devices.
+
+In order to safely constrain such a device, we must interpose something between the device and the memory subsystem.
+This is necessary in order to both limit the damage the device can do, and to protect the integrity of a CHERI system.
+Current systems use an IOMMU (also called a System MMU) to apply a page-based protection and translation model.
+This has significant performance implications due to the highly dynamic behavior of complex devices.
+It may be desirable to augment or replace the IOMMU with some other structure which instead uses the capability model, along the same lines of using CHERI for protection and the MMU for translation.
+Examples of such schemes are given in Cheng and Markettos~\cite{DBLP:conf/micro/MarkettosBBNMW20, cheng2025-capchecker}.
+
+Where possible, a preferable design point is to integrated CHERI support into a bus-mastering device such that the hardware can be trusted to manipulate capabilities correctly.
```
nwf:
```suggestion
Where possible, a preferable design point is to integrate CHERI support into a bus-mastering device such that the hardware can be trusted to manipulate capabilities correctly.
```

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+In this instance we are able to run untrusted I/O software stack safely on trustworthy CHERI-enabled hardware.
```
nwf, line 74: Mention something about attestation?  If we're worried about malicious hot-plug devices claiming to honor CHERI invariants, we need some way to decide whether we believe that.

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+An example of this is a CHERI-enabled GPU~\cite{UCAM-CL-TR-997,naylor26-simtight}, which trustworthily shares capabilities with a CHERI-enabled CPU core.
+
+\subsection{Composing CHERI and DMA}
+DMA controllers operate outside the world of CPU-derived capabilities.
+The question arises how to compose them.
+Below we describe some different levels of utilizing capabilities with DMA.
+
+\subsubsection{Level 0: DMA clears tags}
```
nwf, line 83: This has implications on the software driver model, too: a completely CHERI-oblivious DMA engine grants the authority to access memory given only an integer address, even if it cannot forge capabilities capturing that authority.  As such, access to such DMA engines must be carefully controlled and restricted to TCB components.

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+The simplest yet safe scenario is that DMA operates outside of the world of a CHERI-enabled CPU.
+The DMA hardware is unchanged, using integer address rather than capabilities.
+Any data written by the DMA clears tags, so the DMA is unable to forge or unsafely modify valid capabilities.
+Any pre-existing peripheral device can safely operate in this manner, although this also means they will not follow the capability protection model; an IOMMU may be required to safely constrain their memory access.
+
+\subsubsection{Level 1: addresses replaced by capabiities}
```
nwf, line 90: FWIW, Nazerke Turtayeva did some work on this when she was an intern at MSR Cambridge a few years ago.  Her experiment with the CHERIoT Ibex core is open-source at https://github.com/microsoft/cheriot-ibex/tree/cheriot-dma .

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+A `CHERI-aware DMA controller' may choose to replace integer addresses in descriptors or MMIO registers with capabilities.
+Memory accesses are checked against the bounds and permissions of the capability during transactions, and invalid transactions are aborted.
+Address calculation is based on the address from the capability and is otherwise unchanged.
```
nwf, line 93: To tie some of the text together, this is a particularly simple example of a CHERI-aware bus-mastering peripheral.

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+
+This avoides any difficulties where the DMA controller's use of addresses and the capability model's use of bounds diverge.
+For example, a DMA controller may have separate `offset' and `length' fields to deal with restarting partially-completed transactions (for example, a receive transaction where there was insufficient data received to fill the buffer);
+these would not map directly to the bounds of a capability.
+
+An implementation using DMA using MMIO registers would extend address registers to capabilities, while one using descriptors would need to change the descriptor format to include a capability.
+This may cause downstream effects, such as requiring descriptors to be padded and aligned to multiples of the capability size, which are specific to the DMA controller being modified.
```
nwf, line 105: If descriptors are stored in memory, the DMA engine's access to them would, ideally, also be constrained by capabilities.

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+It is desirable to use the same field for both the length of the transaction as well as its bounds.
+In this way capability manipulation operations in software directly mirror the protection model in hardware, and notably operations loading and storing base and bounds become atomic.
+It can also reduce descriptor size and limit the need for padding.
+This causes some difficulties, however.
+
+First of all, compressed bounds are imprecise.
+This may not be an issue for small buffers, but may become one for larger buffers.
+Additionally alignment may depend on external factors outside our control.
+For example, it may be necessary to restart partial transactions.
+Conceptually it would be possible to increment the address and reduce the bounds, but this could runs into precision difficulties.
```
nwf:
```suggestion
Conceptually it would be possible to increment the address and reduce the bounds, but this could run into precision difficulties.
```

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+Address calculation is based on the address from the capability and is otherwise unchanged.
+
+This avoides any difficulties where the DMA controller's use of addresses and the capability model's use of bounds diverge.
+For example, a DMA controller may have separate `offset' and `length' fields to deal with restarting partially-completed transactions (for example, a receive transaction where there was insufficient data received to fill the buffer);
+these would not map directly to the bounds of a capability.
+
+An implementation using DMA using MMIO registers would extend address registers to capabilities, while one using descriptors would need to change the descriptor format to include a capability.
+This may cause downstream effects, such as requiring descriptors to be padded and aligned to multiples of the capability size, which are specific to the DMA controller being modified.
+
+\subsubsection{Level 2: unifying length and bounds}
```
nwf, line 108: This section conflates "unifying length and bounds" and _re-bounding_ provided capabilities as part of a DMA transaction.  I think those should be pulled apart.

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+There exist a number of potential solutions in this space.
+For a successful safe capability check, we need to also know within which address space a capability exists.
+This leads to roughly three design points.  Either we:
+\begin{enumerate}
+	\item carry around the address space stored as bits inside the capability, or
+	\item maintain the color as external metadata that is supplied to wherever a capability needs to be checked, or
+	\item arrange such that structurally there is no ambiguity at the point of check as to which address space a capability refers.
+\end{enumerate}
+
+In the first case, bits that travel within the in-memory capability are used to identify its address-space identifier (ASID), such as a `color' (see section~\ref{app:exp:peripherals}).
```
nwf, line 162: These intra-capability ASID bits are meant to be interpreted by an IOMMU, SIOV-style, I presume?

app-exp-dma.tex
```diff
@@ -0,0 +1,157 @@
...
+This means that not only are the capabilities with respect to different address spaces, but they represent two different \emph{kinds} of address spaces, potentially whose sources of truth do not overlap.
+
+There exist a number of potential solutions in this space.
+For a successful safe capability check, we need to also know within which address space a capability exists.
+This leads to roughly three design points.  Either we:
+\begin{enumerate}
+	\item carry around the address space stored as bits inside the capability, or
+	\item maintain the color as external metadata that is supplied to wherever a capability needs to be checked, or
+	\item arrange such that structurally there is no ambiguity at the point of check as to which address space a capability refers.
+\end{enumerate}
```
nwf, line 160: It's probably worth discussing the current state(s) of play in non-CHERI systems, especially with PCIe and virtualization-supporting gateware technologies (with fabric-enforced device identifiers and per-device ASID namespaces) and the corresponding software dances to keep IOMMUs and MMUs sufficiently in sync.

app-exp-dma.tex
tmarkettos: Good idea. I've refactored this section into '2 Descriptors; 2a Descriptors fetched from RAM; 2b Descriptors streamed from somewhere else' and put in a comment about how 1.MMIOs may not be organised as descriptors.

app-exp-dma.tex
tmarkettos: I mentioned attestation, as one of various out-of-band ways to find out if you trust the hardware or not

app-exp-dma.tex
tmarkettos: I'm aware of that work, but it doesn't appear to have any documentation I can cite here? It just looks like a big source code drop without much in the way of explanatory comments?

app-exp-dma.tex
tmarkettos: I added a note about that.

app-exp-dma.tex
tmarkettos: I added a mention about that.

app-exp-dma.tex
tmarkettos: I added a note about that - if it's not behind an IOMMU, you need to control it very carefully.

tmarkettos: I mentioned that access must either be externally protected (eg via IOMMU) or constrained to the TCB.

app-exp-dma.tex
tmarkettos: Or something else like a CapChecker. Something has to flatten everything to physical addresses space(s), at which point the colour gives you a clue about which virtual address it relates to.  I added a note to that effect.

app-exp-dma.tex
tmarkettos: I've put in a small paragraph covering that. There is more text on the colouring idea in the subsequent section C.17 so further text might be better there.

app-exp-dma.tex
tmarkettos: I tried to be a bit more explicit here (bounds=length is one thing, then updating that bounds=length is another).

app-exp-dma.tex
qwattash: I agree, this is especially relevant for systems similar to NVIDIA Hopper GPU confidential compute infrastructure.

app-exp-dma.tex
```diff
@@ -0,0 +1,179 @@
...
+Any pre-existing peripheral device can safely operate in this manner, although this also means they will not follow the capability protection model; an IOMMU may be required to safely constrain their memory access.
+Without additional protection, access to such devices must be constrained to within the TCB.
+
+\subsubsection{Level 1: addresses replaced by capabiities}
+A `CHERI-aware DMA controller' may choose to replace integer addresses in descriptors or MMIO registers with capabilities.
+Memory accesses are checked against the bounds and permissions of the capability during transactions, and invalid transactions are aborted.
+Address calculation is based on the address from the capability and is otherwise unchanged.
+Such a design also serves as an example of a simple CHERI-aware bus-mastering peripheral, in that it understands capabilities and imposes their protection model on its transactions.
+
+This avoides any difficulties where the DMA controller's use of addresses and the capability model's use of bounds diverge.
```
qwattash:
```suggestion
This avoids any difficulties where the DMA controller's use of addresses and the capability model's use of bounds diverge.
```

app-exp-dma.tex
```diff
@@ -0,0 +1,179 @@
...
+This may not be an issue for small buffers, but may become one for larger buffers.
+Alignment may depend on external factors outside our control.
+
+Additionary, DMA transactions commonly update descriptors during their course of operation.
+For example, a request for data from an input device such as a UART may not return a full buffer if only a limited amount of data has been received.
+We may wish to update the bounds on the descriptor to reflect that a buffer is partially filled, and then re-queue the descriptor to restart when more data is available.
+Such re-bounding must both reflect both the behavior of the device, and also fit within the capability representability rules.
+Conceptually it would be possible to increment the address and reduce the bounds, but this could run into precision difficulties.
+Consider a 1~MiB-sized buffer send in which one byte has been successfully transferred so far.
+We would wish to increment the address by one and set the bounds to 1~MiB minus 1 but that size may be unrepresentable under the capability compression scheme.
```
qwattash, line 124: It is unclear to me how much of a problem this may be. If more data is not appended to the buffer, there is a question of what uninitialized data lays in the representability padding. This means that there are further considerations to be made with respect with the temporal safety of DMA buffers, which I believe is mostly outside the control of the DMA engine / devices.

app-exp-dma.tex
```diff
@@ -0,0 +1,179 @@
...
+In this instance we are able to run untrusted I/O software stack safely on trustworthy CHERI-enabled hardware.
+An example of this is a CHERI-enabled GPU~\cite{UCAM-CL-TR-997,naylor26-simtight}, which trustworthily shares capabilities with a CHERI-enabled CPU core.
+In this situation we need some external assurance of hardware trustworthiness, either by construction (e.g. purchase of silicon IP from a trusted vendor) or at composition time (e.g. a hotplugged device can cryptographically attest it is from a trusted vendor or of a trustworthy design).
+
+\subsection{Composing CHERI and DMA}
+DMA controllers operate outside the world of CPU-derived capabilities.
+The question arises how to compose them.
+Below we describe some different levels of utilizing capabilities with DMA.
+
+\subsubsection{Level 0: DMA clears tags}
```
qwattash, line 83: It may not be obvious to the reader here that capabilities refer to the physical address space (unless I'm mistaken), since most of the CHERI document deals with virtually addressed capabilities. This would also tie with the "Level 3" virtual addressing discussion. Perhaps it is worth cross referencing the physical caps experimental section?

app-exp-dma.tex
```diff
@@ -0,0 +1,179 @@
...
+In the second, the in-memory format is not changed but bits are supplied alongside, `out of band' (e.g. as additional tag bits).
+In the third, we may use explicit or implicit information to infer the address space.
+For example, we know a particular interconnect endpoint always generates capabilities within a single address-space, allowing us to attach an ASID at the point of check.
+An example of this inference (with respect to an object ID rather than an ASID) is described in Cheng~\cite{cheng2025-capchecker}.
+In each case we would need to ensure any manipulation of capabilities preserves these additional properties.
+
+When checking an access against the capability we must both check the operation is permitted by the capability (i.e. the usual access check), and that also that the access is permitted by rules associated with the ASID.
+For example, deriving a new capability must retain the ASID of the original one (no implicit type conversion);
+any operation using two capabilities of different ASIDs is not permitted (no type mixing);
+and capabilities are not allowed to be stored in locations associated with another ASID (no leakage).
```
qwattash, line 172: I wonder whether the ASID in this case can be somehow mapped to a multi-bit Capability Level model.

app-exp-dma.tex
tmarkettos: The general point is that DMA subsystems may use the length field in a way that doesn't fit the capability model. Representability is one potential issue, another is doing surprising things like increasing the length for some reason.  I'm also unclear how much of a problem this is as data on this is likely to be difficult to obtain.

I think the higher-level point is that bounds and length don't mean the same thing, and the points where they differ may cause friction.  The exact manifestation of friction may vary, and it's up to the designer to decide whether they can adapt their system to make them the same, or if there's an annoying corner case where they need to diverge.  It's really up to them to work through their specific situation.

app-exp-dma.tex
tmarkettos: Thanks, I added a paragraph hopefully clarifying the situation.

app-exp-dma.tex
tmarkettos: You can use multiple tag bits to represent the ASID, but it depends what rules you have for using multiple tag bits with a multi-level capability model.  eg if you were doing the classic multilevel security thing then it wouldn't be representing the same thing (unless the hardware was also following MLS, eg the classified plaintext NIC and the unclassified ciphertext NIC). How you store the ASID and manage it is a design choice you have to make.

app-exp-dma.tex
davidchisnall: There's an entire family of PCIe specifications surrounding this.

tmarkettos:
I've commented and updated various parts of the PR, but it seems none of my comments have been submitted.  Let me see if posting this stops them being 'pending'...

tmarkettos (commented):
I've made numerous comments which Github is hiding. Let me see if this submits them now...

Edit: it did

